PR #16, opened 2021-01-07
Fix to respect dependsOn field

### LINKED ISSUE #13: Must respect DependsOn property of Workloads
Currently, we're not respecting `Workload.DependsOn`. Internally, it is used to ensure workloads whose installation depends on the installation status of others' are installed in the right order.

For 10.0.17, a change was introduced that makes this bug crucial to fix; without it, scale unit workload installation will often fail with: "Missing setup records, please make sure that hub's first export was successful.", as SYS is not guaranteed to be installed before {WES/MES}.

alkalomo:
closes #13

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
@@ -1,10 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Threading.Tasks;
 using CloudAndEdgeLibs.AOS;
 using CloudAndEdgeLibs.Contracts;
-using DeepEqual.Syntax;
-using FluentAssertions;
 using ScaleUnitManagement.Utilities;
 using ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities;
 
@@ -41,7 +40,7 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator
 
                 foreach (WorkloadInstanceIdWithName workloadInstanceIdWithName in workloadInstanceIdWithNameList)
                 {
-                    statusList.Add(await scaleUnitAosClient.CheckWorkloadStatus(workloadInstanceIdWithName.WorkloadInstanceId));
+                    statusList.Add(await WorkloadInstanceManager.GetWorkloadInstanceStatus(scaleUnitAosClient, workloadInstanceIdWithName.WorkloadInstanceId));
                 }
                 foreach (WorkloadInstanceStatus status in statusList)
                 {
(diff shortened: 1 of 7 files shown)


src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceTopologicalSortUtil.cs
```diff
@@ -0,0 +1,102 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using CloudAndEdgeLibs.Contracts;
+
+namespace ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities
+{
+    class WorkloadInstanceTopologicalSortUtil
```
AndreasHassing, line 8: Can you add a brief comment about what this is for (explaining the DependsOn concept, not very detailed).

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -51,34 +50,59 @@ public async Task InstallationStatus(int input, string selectionHistory)
...
+                return;
 
-            await ReliableRun.Execute(async () =>
+            Console.WriteLine($"Waiting for the {workloadInstance.VersionedWorkload.Workload.Name} workload initial sync to complete");
+
+            int count = 0;
+            do
             {
-                AOSClient hubAosClient = await AOSClient.Construct(Config.HubAosResourceId(), Config.HubAosEndpoint());
+                for (int i = 0; i < 10; i++) // wait 10 seconds before querying the status again.
```
AndreasHassing, line 63: Let's bring it down a little. We can wait 3-5 seconds.

AndreasHassing (changes requested):
LGTM 👍 - thanks for this.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
alkalomo: Based on my experience, it takes a couple of minutes to get each workload installed, and it felt just right. I'm going to leave it as is for now.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceTopologicalSortUtil.cs
alkalomo: sure

src/ScaleUnitManagementTests/WorkloadInstanceTopologicalSortUtilTest.cs
```diff
@@ -0,0 +1,166 @@
...
+            // Act
+            WorkloadInstanceTopologicalSortUtil topologicalSortUtil = new WorkloadInstanceTopologicalSortUtil(workloadInstances);
+            List<WorkloadInstance> sortedWorkloadInstances = topologicalSortUtil.Sort();
+
+            // Assert
+            List<WorkloadInstance> expectedSortedList = new List<WorkloadInstance> { w4, w2, w1, w3 };
+            
+            for (int i = 0; i < sortedWorkloadInstances.Count; i++)
+            {
+                sortedWorkloadInstances[i].Should().Be(expectedSortedList[i]);
```
AndreasHassing, line 145: [nit] I think you can just do `sortedWorkloadInstances.Should().Equal(expectedSortedList)`

AndreasHassing (approved):

PR #17, opened 2021-01-08
enable step discover from other assemblies

alkalomo:


src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/StepFactory.cs
```diff
@@ -1,16 +1,25 @@
...
 {
     public class StepFactory
     {
-        private IEnumerable<Type> FindClassesDeriving(Type type)
+        private List<Type> FindClassesDeriving(Type type)
         {
-            return (Assembly.GetAssembly(type)).GetTypes().Where(TheType => TheType.IsClass && !TheType.IsAbstract && TheType.IsSubclassOf(type));
+            List<Type> res = new List<Type>();
+
+            foreach (Assembly a in AppDomain.CurrentDomain.GetAssemblies())
```
AndreasHassing, line 14: Would a LINQ expression like the following work?
```csharp
return
    from assembly in AppDomain.CurrentDomain.GetAssemblies()
    from type in assembly.GetTypes()
    where type.IsClass && !type.IsAbstract && type.IsSubClassOf(superType)
    select type;
```

AndreasHassing (approved):

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/StepFactory.cs
alkalomo: yes indeed.

AndreasHassing (approved):
LGTM! 👍

PR #18, opened 2021-01-08
Add testing to PR workflow

AndreasHassing:
@alkalomo recently added test to the repo, we should run those as part of the PR workflow check.


alkalomo (approved):
Looks good

PR #22, opened 2021-01-31
remove spoke from user config and remove unnecessary dependency that leads to build issues.

alkalomo:
remove spoke from user config and remove unnecessary dependency that leads to build issues.

### DIFF
--- a/src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/IISAdministrationHelper.cs
+++ b/src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/IISAdministrationHelper.cs
@@ -12,11 +12,12 @@ namespace ScaleUnitManagement.ScaleUnitFeatureManager.Utilities
         internal static void CreateSite(string siteName, string siteRoot,
             string bindingInformation, string certSubject, string appPoolName)
         {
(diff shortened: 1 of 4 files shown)


src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/IISAdministrationHelper.cs
```diff
@@ -12,11 +12,12 @@ internal static class IISAdministrationHelper
         internal static void CreateSite(string siteName, string siteRoot,
             string bindingInformation, string certSubject, string appPoolName)
         {
-            Requires.NotNull(siteName, nameof(siteName));
-            Requires.NotNull(siteRoot, nameof(siteRoot));
-            Requires.NotNull(bindingInformation, nameof(bindingInformation));
-            Requires.NotNull(certSubject, nameof(certSubject));
-            Requires.NotNull(appPoolName, nameof(appPoolName));
+            if (String.IsNullOrEmpty(siteName)
```
AndreasHassing, line 15: Ideally we want to know which argument is missing to help diagnose future issues.

AndreasHassing (approved):

Issue #30, opened 2021-02-04
Emit errors hidden behind retries


AndreasHassing:
Currently, errors caught in a retry-mode operation are completely hidden from the user. They should at least be visible in a file somewhere so the user can go and read them while the program retries or after-the-fact.

Issue #29, opened 2021-02-04
Should throw configuration error when hub and scale unit IP are the same


AndreasHassing:
As same-IP bindings cause problems in IIS when using a shared-box hub and scale unit.

Issue #27, opened 2021-02-04
Tool attempts to uninstall/stop scale unit Batch service if it does not exist


AndreasHassing:
This is minor; the tool currently shows an error while enabling the hybrid topology on a shared-box scale unit (when it creates the additional Batch service) - the error is benign as we ignore it, but we can avoid it completely if we just check if the Batch service is installed or not.

gammelgaard52:
Hi,
It seems like challenge is still there.
![image](https://user-images.githubusercontent.com/16650923/107034199-ae3eeb80-67b6-11eb-849e-ef5a60a54e79.png)

Scenario:
No service installed
Repro:
Run cli.exe, select 1, select 1, select 1 (stop service)

AndreasHassing:
Thanks @gammelgaard52, I missed that one 😊.

Issue #25, opened 2021-02-04
Enable the MES flight during the initialization of the hybrid topology


alkalomo:
The flight that needs to be enabled is SysWorkloadTypeMESFeatureToggle.

PR #31, opened 2021-02-04
Assorted quality-of-life improvements

AndreasHassing:
Closes the following issues:
- #25 
- #23 
- #27 
- #29 
- #30

### DIFF
--- a/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddMESFeatureFlight.cs
+++ b/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddMESFeatureFlight.cs
@@ -14,7 +14,8 @@ namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
 
         public float Priority()
         {
-            // before starting services again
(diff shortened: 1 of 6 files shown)


src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddMESFeatureFlight.cs
```diff
@@ -0,0 +1,38 @@
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
+using ScaleUnitManagement.Utilities;
+
+namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
+{
+    public sealed class AddMESFeatureFlight : ICommonStep
```
alkalomo, line 6: This should only run on the hub I think

alkalomo (approved):

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddMESFeatureFlight.cs
AndreasHassing: Discussed offline :-).

PR #33, opened 2021-02-05
"Admin required" check when tool starts

cabeln:


src/CLI/Program.cs
```diff
@@ -1,22 +1,47 @@
...
+            var workloadsInstallationStatusOption = new CLIOption() { Name = "Show workloads installation status", Command = WorkloadsInstallationStatus.Show };
 
             var options = new List<CLIOption>() { enableScaleUnitFeatureOption, configureEnvironmentOption, installWorkloadsOption, workloadsInstallationStatusOption };
 
-            CLIScreen screen = new CLIScreen(options, "Home", "Please select the operation you want to perform:\n", "\nOperation to perform: ");
+            var screen = new CLIScreen(options, "Home", "Please select the operation you want to perform:\n", "\nOperation to perform: ");
             await CLIMenu.ShowScreen(screen);
         }
+
+        private static bool HasAdministratorPrivileges()
```
alkalomo, line 40: Can we place this in some other class? We already have this 
https://github.com/microsoft/SCMScaleUnitDevTools/blob/main/src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/CheckForAdminAccess.cs

Can we reuse it? Maybe we need to update the code there.

src/CLI/Program.cs
```diff
@@ -1,22 +1,47 @@
...
-            CLIOption installWorkloadsOption = new CLIOption() { Name = "Install workloads", Command = InstallWorkloads.Show };
-            CLIOption workloadsInstallationStatusOption = new CLIOption() { Name = "Show workloads installation status", Command = WorkloadsInstallationStatus.Show };
+            if (!HasAdministratorPrivileges())
+            {
+                Console.Error.WriteLine("Access Denied!\r\n" +
+                    "This tool must be run with administrator permissions.\r\n" +
+                    "Please start the tool in an administrator command prompt.\r\n\r\n" + 
+                    "Press any key to exit. . .");
+                Console.ReadKey();
+                return 740; // ERROR_ELEVATION_REQUIRED
```
alkalomo, line 20: Let's take this out in a method called something like ValidateProcessHasAdministratorPrivileges() then the main method can stay as is with a call to this new method.

src/CLI/Program.cs
```diff
@@ -1,22 +1,50 @@
...
 using System.Threading.Tasks;
 using CLIFramework;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
 
 namespace CLI
 {
     public class Program
     {
-        public static async Task Main(string[] args)
+        public static int Main(string[] args)
```
alkalomo, line 12: sorry for not making this clear before. Changing the signature of the Main method, is a breaking change (will break our internal repo).

We can avoid the breaking change by keeping the Main method as is
```
public static async Task Main(string[] args)
{
CheckForAdminAccess.ValidateCurrentUserIsProcessAdmin(); // this can simply exit the process using Environment.Exit()
// code that generates the main screen
}
```

src/CLI/Program.cs
```diff
@@ -1,22 +1,50 @@
...
-            CLIOption installWorkloadsOption = new CLIOption() { Name = "Install workloads", Command = InstallWorkloads.Show };
-            CLIOption workloadsInstallationStatusOption = new CLIOption() { Name = "Show workloads installation status", Command = WorkloadsInstallationStatus.Show };
+            try
+            {
+                CheckForAdminAccess.ValidateCurrentUserIsProcessAdmin();
+            }
+            catch (Exception ex)
+            {
+                Console.Error.WriteLine(ex.Message);
+                Console.Error.WriteLine("\r\n\r\nPress any key to exit. . .");
```
alkalomo, line 21: This should be moved inside the ValidateCurrentUserIsProcessAdmin()

src/CLI/Program.cs
```diff
@@ -1,22 +1,50 @@
...
+            var workloadsInstallationStatusOption = new CLIOption() { Name = "Show workloads installation status", Command = WorkloadsInstallationStatus.Show };
 
             var options = new List<CLIOption>() { enableScaleUnitFeatureOption, configureEnvironmentOption, installWorkloadsOption, workloadsInstallationStatusOption };
 
-            CLIScreen screen = new CLIScreen(options, "Home", "Please select the operation you want to perform:\n", "\nOperation to perform: ");
+            var screen = new CLIScreen(options, "Home", "Please select the operation you want to perform:\n", "\nOperation to perform: ");
             await CLIMenu.ShowScreen(screen);
         }
+
+        private static bool HasAdministratorPrivileges()
```
alkalomo, line 43: If this is not used, then remove

src/CLI/Program.cs
```diff
@@ -1,22 +1,50 @@
+using System;
 using System.Collections.Generic;
+using System.Security.Principal;
```
alkalomo, line 3: let's remove what is not needed.

alkalomo (changes requested):

alkalomo (approved):

PR #44, opened 2021-03-08
Remove AAD Tenant ID from user config

### LINKED ISSUE #43: Remove TenantId from UserConfig
Edit: Modifying the TenantId solves a very specific pain internally, but it turned out to create more problems than it solved. We have solved our issues at the AAD infrastructure level and can now remove the field entirely.

~~Currently, we're only changing the `Aad.AADTenantId` based on what the user provides in [`Config.xml`](https://github.com/microsoft/SCMScaleUnitDevTools/blob/main/src/ScaleUnitManagement/UserConfig.sample.xml). If this does in fact change the tenant while `Aad.TenantDomainGUID` is not changed, this can cause authentication issues (seemingly especially when using PPE AAD authority).~~

~~To avoid needing an additional element in the `UserConfig` file, we can collect the tenant object id using `Aad.AADTenantId` and set `Aad.TenantDomainGUID` to this value to ensure consistency and avoid authentication errors.~~

~~Something like this ought to work:~~
```csharp
AuthenticationResult authenticationResult = await OAuthHelper.GetAuthenticationResult("https://login.windows.net/<tenant domain>", "<app id>", "<app secret>", "https://usnconeboxax1aos.cloud.onebox.dynamics.com");

JwtSecurityToken token = new JwtSecurityTokenHandler().ReadJwtToken(authenticationResult.AccessToken);

// the `tid` claim contains the tenant domain GUID
string? tidClaimValue = token.Claims.FirstOrDefault(c => c.Type.Equals("tid", StringComparison.Ordinal))?.Value;

Console.WriteLine(tidClaim is null ? "Did not find `tid` claim" : $"[tid]: {tidClaimValue}");

// now we can inject `tidClaimValue` into `Aad.TenantDomainGUID` in web.config
```

Sample error message from having `Aad.AADTenantId` and `Aad.TenantDomainGUID` mismatch:
```
Hub configuration failed. Retry count exceeded. Execution will not continue.
Unhandled exception. System.Exception: [401]: Error - {"Message":"Authentication failed.","StackTrace":null,"ExceptionType":"System.InvalidOperationException"}
```

AndreasHassing:
Closes #43.


src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddMESFeatureFlight.cs
```diff
@@ -0,0 +1,41 @@
+using System.Threading.Tasks;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
+using ScaleUnitManagement.Utilities;
+
+namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
```
alkalomo, line 5: Is this right? I think this probably came back after merging?

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/SharedWebConfig.cs
```diff
@@ -0,0 +1,43 @@
...
+                    try
+                    {
+                        string tenantDomainGUID = await OAuthHelper.GetTidClaim(Config.AADTenantId(), Config.AppId(), Config.AppSecret(), Config.HubScaleUnit().ResourceId());
+
+                        if (!string.IsNullOrWhiteSpace(tenantDomainGUID))
+                            webConfig.UpdateXElement("Aad.TenantDomainGUID", tenantDomainGUID);
+                    }
+                    catch (Exception ex)
+                    {
+                        Console.Error.WriteLine($"Unable to get tenant ID for {Config.AADTenantId()}, skipping setting Aad.AADTenantDomainGUID in Web.config:\n{ex}");
```
alkalomo, line 30: If this fails should we continue? We can't use the tool right?

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddMESFeatureFlight.cs
AndreasHassing: That's a mistake, thanks!

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/SharedWebConfig.cs
AndreasHassing: It could be that AAD is unavailable but we already updated the value previously. I'm not sure.

Hmm.. I think you're right, we should fail.

alkalomo (approved):

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/WebConfigAudienceConfigurator.cs
```diff
@@ -0,0 +1,36 @@
+using System.Collections.Generic;
+using System.Linq;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
+using ScaleUnitManagement.Utilities;
+
+namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
+{
+    public static class WebConfigAudienceConfigurator
+    {
+        public static void AddValidAudiences(WebConfig webConfig)
```
alkalomo, line 10: Would it make sense to put this as a method in the webconfig class? We already have an AddKey there, to me it's of similar nature.

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -29,10 +29,25 @@ private static CloudAndEdgeConfiguration UserConfiguration()
         public static bool UseSingleOneBox() { return UserConfiguration().UseSingleOneBox; }
         public static string AppId() { return UserConfiguration().AADConfiguration.AppId; }
         public static string AppSecret() { return UserConfiguration().AADConfiguration.AppSecret; }
+        public static string AppResourceId(ScaleUnitInstance scaleUnitInstance = null)
+        {
+            var userConfig = UserConfiguration();
+
+            return string.IsNullOrWhiteSpace(userConfig.AADConfiguration.AppResourceId)
```
alkalomo, line 36: is this going to catch the case where we have the node defined but with no value?

alkalomo (approved):

PR #46, opened 2021-03-19
Admin -> ScaleUnitManagement.

DmytroPohorielov:
ScaleUnitManagement user is the user to be created during the AX installation/upgrade. It is assigned a limited role just enough to execute the setup routines.

### DIFF
--- a/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddToolClientToSysAADClientTable.cs
+++ b/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddToolClientToSysAADClientTable.cs
@@ -23,7 +23,7 @@ namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
 USE {scaleUnit.AxDbName};
 
 IF NOT EXISTS (SELECT TOP 1 1 FROM SysAADClientTable WHERE AADClientId = '{scaleUnit.AuthConfiguration.AppId}')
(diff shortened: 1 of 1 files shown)


src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/AddToolClientToSysAADClientTable.cs
```diff
@@ -23,7 +23,7 @@ public Task Run()
 USE {scaleUnit.AxDbName};
 
 IF NOT EXISTS (SELECT TOP 1 1 FROM SysAADClientTable WHERE AADClientId = '{scaleUnit.AuthConfiguration.AppId}')
-    INSERT INTO SysAADClientTable (AADClientId, UserId, Name) VALUES ('{scaleUnit.AuthConfiguration.AppId}', 'Admin', 'Scale Unit Management Tool');
+    INSERT INTO SysAADClientTable (AADClientId, UserId, Name) VALUES ('{scaleUnit.AuthConfiguration.AppId}', 'ScaleUnitManagement', 'Scale Unit Management Tool');
```
AndreasHassing, line 26: Until the role is found in the deployed code, we should keep using `Admin`.

Can you add a runtime check with [`AxDeployment.IsApplicationVersionMoreRecentThan("<build number where your changes went into Application git>")`](https://github.com/microsoft/SCMScaleUnitDevTools/blob/4cb28d62adb80e3a2e8f77837b755fda3d083818/src/ScaleUnitManagement/Utilities/AxDeployment.cs#L37) here to branch on use of either of the two users.

AndreasHassing (changes requested):

AndreasHassing (approved):
LGTM 👍.

PR #64, opened 2021-07-20
Set temporal assignments' effective date in the future =< 10.0.21

Antonstockmarr:
Changing the effective date on the workloads when they are installed.
If the version is 10.0.21 or earlier, 5 minutes are added to the time, such that the installation succeeds first and the movement jobs will get created.

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceManager.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceManager.cs
@@ -66,7 +66,7 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities
                             {
                                 new TemporalAssignment()
                                 {
-                                    EffectiveDate = DateTime.UtcNow,
+                                    EffectiveDate = GetWorkloadEffectiveDate(),
                                     Environment = new PhysicalEnvironmentReference()
                                     {
                                         Id = Config.ScaleUnitEnvironmentId,
@@ -248,7 +248,7 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities
                     {
                         new TemporalAssignment()
                         {
-                            EffectiveDate = DateTime.UtcNow,
+                            EffectiveDate = GetWorkloadEffectiveDate(),
                             Environment = new PhysicalEnvironmentReference()
                             {
                                 Id = Config.ScaleUnitEnvironmentId,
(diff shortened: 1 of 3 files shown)


src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceManager.cs
```diff
@@ -276,6 +276,16 @@ private List<WorkloadInstance> GetSYSWorkloadInstancesPerLegalEntity(Workload wo
             return sysWorkloadInstances;
         }
 
+        private static DateTime GetWorkloadEffectiveDate()
+        {
+            DateTime effectiveDate = DateTime.UtcNow;
+            if (!AxDeployment.IsApplicationVersionMoreRecentThan("10.0.21"))
```
AndreasHassing, line 282: Unfortunately, build versions are not mapped 1-to-1 with Application Release version names (i.e. 10.0.21 won't have build version 10.0.21.<patch>). This is something I forgot to mention, so this is a fair attempt 😊.

For finding deployments on 10.0.22 or higher, you can use the following build number: `10.13.0`, which implies 10.0.22 or higher.

10.0.21 build number: `10.12.<patch>`.

AndreasHassing (changes requested):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceManager.cs
AndreasHassing: Consider adding a test to the test which targets `AxDeployment`-util method `IsApplicationVersionMoreRecentThan` which ensures that a checked version number `10.13.0` returns true for when the deployed application version number is `10.13.1837`.

AndreasHassing:
This PR will close #65.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceManager.cs
Antonstockmarr: So should I create a separate test method for `IsApplicationVersionMoreRecentThan` that specifically tests versions `10.13.0` and `10.13.1837`?
And, just to confirm, a checked version number `10.13.0` is not more recent than `10.13.1837`, so it should return false, right?

AndreasHassing: Yes, also consider testing that `10.13` works as well (so we can remove the `0` altogether and convey that we don't care which patch version it is).

For the checked number, semantically it's the other way around: we want to check that the current application version (could be `10.13.1837`, should always have a patch number) is more recent than, say, `10.13.0` and `10.13`.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/WorkloadInstanceManager.cs
```diff
@@ -276,6 +276,16 @@ private List<WorkloadInstance> GetSYSWorkloadInstancesPerLegalEntity(Workload wo
             return sysWorkloadInstances;
         }
 
+        private static DateTime GetWorkloadEffectiveDate()
+        {
+            DateTime effectiveDate = DateTime.UtcNow;
+            if (!AxDeployment.IsApplicationVersionMoreRecentThan("10.13"))
```
AndreasHassing, line 282: Consider adding a comment here indicating what `10.13` means, or use a variable for the same effect; something like:

```csharp
string app10_0_22BaseVersion = "10.13";
```

AndreasHassing (approved):
LGTM! Added a single comment.

PR #66, opened 2021-07-22
Checking if the scaleUnitId is valid

### LINKED ISSUE #63: Validate if ScaleUnitId is valid
Currently, anything can be placed in the `<ScaleUnitId>` field in `UserConfig.xml`. We should validate that the ID is within the mnemonic range (i.e. between @@ (0) and 1000).

Antonstockmarr:
// Not tested

Checking when the config is loaded if the ScaleUnitId is valid, i.e. if it lies within 1 and 1000 or is a hub.

resolves #63

### DIFF
--- a/src/ScaleUnitManagement/Utilities/Config.cs
+++ b/src/ScaleUnitManagement/Utilities/Config.cs
@@ -153,6 +153,7 @@ namespace ScaleUnitManagement.Utilities
                 if (!scaleUnit.IsHub() && (integralScaleUnitId < 1 || 1000 <= integralScaleUnitId))
                     Console.Error.WriteLine("ScaleUnitId is not in the mnemonic range (between 0 and 1000)");
 
+
                 if (Config.ScaleUnitInstances().Where(s => s.ScaleUnitId == scaleUnit.ScaleUnitId).Count() != 1)
                     Console.Error.WriteLine("ScaleUnitId is not unique");
 
--- a/src/ScaleUnitManagement/Utilities/Config.cs
+++ b/src/ScaleUnitManagement/Utilities/Config.cs
@@ -149,6 +149,10 @@ namespace ScaleUnitManagement.Utilities
                 ValidateValue("AppSecret", scaleUnit.AuthConfiguration.AppSecret);
                 ValidateValue("Authority", scaleUnit.AuthConfiguration.Authority);
 
+                int integralScaleUnitId = ScaleUnitMnemonicCalculator.ToIntegralValue(scaleUnit.ScaleUnitId);
+                if (!scaleUnit.IsHub() && (integralScaleUnitId < 1 || 1000 <= integralScaleUnitId))
+                    Console.Error.WriteLine("ScaleUnitId is not in the mnemonic range (between 0 and 1000)");
+
                 if (Config.ScaleUnitInstances().Where(s => s.ScaleUnitId == scaleUnit.ScaleUnitId).Count() != 1)
                     Console.Error.WriteLine("ScaleUnitId is not unique");
 
--- a/src/ScaleUnitManagement/Utilities/Config.cs
+++ b/src/ScaleUnitManagement/Utilities/Config.cs
@@ -156,7 +156,7 @@ namespace ScaleUnitManagement.Utilities
                         int integralScaleUnitId = ScaleUnitMnemonicCalculator.ToIntegralValue(scaleUnit.ScaleUnitId);
                         if (integralScaleUnitId < 1 || integralScaleUnitId >= 1000)
                         {
-                            Console.Error.WriteLine("ScaleUnitId is not in the mnemonic range (between 1 and 1000)");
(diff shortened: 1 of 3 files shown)


src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -149,6 +149,11 @@ void ValidateValue(string fieldName, string fieldValue)
                 ValidateValue("AppSecret", scaleUnit.AuthConfiguration.AppSecret);
                 ValidateValue("Authority", scaleUnit.AuthConfiguration.Authority);
 
+                int integralScaleUnitId = ScaleUnitMnemonicCalculator.ToIntegralValue(scaleUnit.ScaleUnitId);
+                if (!scaleUnit.IsHub() && (integralScaleUnitId < 1 || 1000 <= integralScaleUnitId))
```
Omaaarz, line 153: It's always better to have the constant you are comparing to on the right side of the operator. i.e 
integralScaleUnitId >= 1000

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -149,6 +149,11 @@ void ValidateValue(string fieldName, string fieldValue)
...
+                    Console.Error.WriteLine("ScaleUnitId is not in the mnemonic range (between 0 and 1000)");
```
Omaaarz, line 154: You will also need to set hasAnyError to true to exit at the end of the method.

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -149,6 +149,11 @@ void ValidateValue(string fieldName, string fieldValue)
                 ValidateValue("AppSecret", scaleUnit.AuthConfiguration.AppSecret);
                 ValidateValue("Authority", scaleUnit.AuthConfiguration.Authority);
 
+                int integralScaleUnitId = ScaleUnitMnemonicCalculator.ToIntegralValue(scaleUnit.ScaleUnitId);
```
Omaaarz, line 152: Maybe we need to catch exceptions from here and write the exception message to the console and set hasAnyError to true if any exceptions found.

Omaaarz: Actually shouldn't we just get the integral value if only it is hub?

src/ScaleUnitManagement/Utilities/ScaleUnitMnemonicCalculator.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            StringBuilder mnemonicString = new StringBuilder();
+            const char baseChar = '@';
+            const int numberBase = 27;
+            while (integralValue > 0)
+            {
+                int remainder = integralValue % numberBase;
+                integralValue /= numberBase;
+                mnemonicString.Append((char)(baseChar + remainder));
```
Omaaarz, line 32: The string size to be returned is very small, so using strings instead of stringbuilder to concatenate won't cause performance issues and it will make the code more readable.

Omaaarz: You can append as following

string += char;

src/ScaleUnitManagement/Utilities/ScaleUnitMnemonicCalculator.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public static int ToIntegralValue(string mnemonic)
+        {
+            mnemonic = mnemonic.ToUpperInvariant();
+
+            // Use '@' as base so that 'A' isn't logically zero, making 'AAA' is different from 'AA', e.g.
+            const char baseChar = '@';
+            int integralValue = 0;
+            int i = 0;
+            foreach (var charIt in mnemonic)
```
Omaaarz, line 51: Since we are defining an int iterator anyway. I prefer using the regular for block
for (int i = 0; i < mnemonic.Length; i++)
{
    int value = mnemonic[i] - baseChar;

Omaaarz:
Can you please add tests for the calculator class?

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -149,9 +149,22 @@ void ValidateValue(string fieldName, string fieldValue)
...
-                int integralScaleUnitId = ScaleUnitMnemonicCalculator.ToIntegralValue(scaleUnit.ScaleUnitId);
-                if (!scaleUnit.IsHub() && (integralScaleUnitId < 1 || 1000 <= integralScaleUnitId))
-                    Console.Error.WriteLine("ScaleUnitId is not in the mnemonic range (between 0 and 1000)");
+                if (!scaleUnit.IsHub())
+                {
+                    try
+                    {
+                        int integralScaleUnitId = ScaleUnitMnemonicCalculator.ToIntegralValue(scaleUnit.ScaleUnitId);
+                        if (integralScaleUnitId < 1 || integralScaleUnitId >= 1000)
+                        {
```
Omaaarz, line 158: The values in error message and in the if condition don't seem consistent?

Omaaarz:
LGTM. Added 1 more minor comment

Omaaarz (approved):

Issue #35, opened 2021-02-08
Provide capability to perform DB sync against project/model for the scale unit DB


alkalomo:
This applies to OneBox only environments, where DBsync will go against the main (hub) environment's db.

AndreasHassing:
This is for targeting the Scale Unit DB with DbSync when `UseSingleOneBox == true` (i.e. both Hub and Scale Unit is found on the same machine), where VS makes it easy to sync single models and perform full sync against the default DB, there is nothing such for the Scale Unit DB.

This functionality should be added to a new menu, unrelated to the scale unit setup itself, but rather a toolbox of QoL tools.

Omaaarz:
closed by PR #67

PR #67, opened 2021-07-26
Creating Setup Tools menu, which allows syncing both the Hub and the Spoke

Antonstockmarr:
Added a new option to the start menu: "Setup Tools". From here the user can sync the HubDB, or select a ScaleUnit to sync. The scale units only appear if UseSingleOneBox is true.

### DIFF
--- a/src/CLI/SetupTools.cs
+++ b/src/CLI/SetupTools.cs
@@ -12,7 +12,6 @@ namespace CLI
 
             options.Add(new CLIOption() { Name = "Sync DB", Command = SyncDB.Show });
 
-
             var screen = new CLIScreen(options, "Home", "Please select the operation you want to perform:\n", "\nOperation to perform: ");
             await CLIMenu.ShowScreen(screen);
         }
--- a/src/CLI/SetupTools.cs
+++ b/src/CLI/SetupTools.cs
@@ -1,0 +1,20 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CLIFramework;
+
+namespace CLI
+{
+    class SetupTools
+    {
+        public static async Task Show(int input, string selectionHistory)
+        {
+            var options = new List<CLIOption>();
+
+            options.Add(new CLIOption() { Name = "Sync DB", Command = SyncDB.Show });
+
+
+            var screen = new CLIScreen(options, "Home", "Please select the operation you want to perform:\n", "\nOperation to perform: ");
+            await CLIMenu.ShowScreen(screen);
+        }
+    }
+}
--- a/src/CLI/SyncDB.cs
+++ b/src/CLI/SyncDB.cs
@@ -13,7 +13,6 @@ namespace CLI
         {
             var options = new List<CLIOption>();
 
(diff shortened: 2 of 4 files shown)


Antonstockmarr:
Resolves #35

src/CLI/SyncDB.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+
+namespace CLI
+{
+    class SyncDB
+    {
+        public static async Task Show(int input, string selectionHistory)
+        {
+            var options = new List<CLIOption>();
+
```
Omaaarz, line 16: nit: extra empty non needed line

src/CLI/SyncDB.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (Config.UseSingleOneBox())
```
Omaaarz, line 18: Why are we branching here?

I guess we can always do what is in the current if body.

src/CLI/SyncDB.cs
```diff
@@ -0,0 +1,67 @@
...
+            else
+            {
+                options.Add(new CLIOption() { Name = "Hub", Command = SyncHubDB });
+            }
+
+            var screen = new CLIScreen(options, "Home", "Please select the database you would like to sync:\n", "\nDatabase to sync: ");
+            await CLIMenu.ShowScreen(screen);
+        }
+
+        private static async Task SyncHubDB(int input, string selectionHistory)
```
Omaaarz, line 37: No need for different methods for hub and spoke DBsync. You can always use the implementation of SyncSpokeDB method but you might need to rename it to be generic (ex: syncDB)

Omaaarz (changes requested):
Left some comments to address

src/CLI/SyncDB.cs
Antonstockmarr: What if a user tries to sync a spoke that is not running on the same machine as the hub? Can that happen, and would the DBsync still work as expected?

Omaaarz: If the DB isn't existing in the current machine, then DB sync would fail. If both envs DBs have the same name, then it would sync the hub db instead (I guess that's still not a problem). 

But looking at what you providing, then in non single box setup, there will be no way to sync spokes(scale units). You will only sync hub which isn't correct.

It will be the responsibility of the developer that runs the operation in the correct environment. I guess we are doing the same for (initialize hybrid topology) step.

Antonstockmarr: Okay, I see what you mean. The user cannot sync the spoke in non-singleBox setup. I will correct it.

src/CLI/SetupTools.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace CLI
+{
+    class SetupTools
+    {
+        public static async Task Show(int input, string selectionHistory)
+        {
+            var options = new List<CLIOption>();
+
+            options.Add(new CLIOption() { Name = "Sync DB", Command = SyncDB.Show });
+
```
Omaaarz, line 14: nit: empty line

Omaaarz (approved):
LGTM. left 1 minor comment.

PR #68, opened 2021-08-10
Move and delete workloads

Antonstockmarr:
Supporting movement of workloads from scale units back to the hub, showing status of workload movements and deleting workloads from a scale unit. All three items have explicit menu options.


src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/AOSClient.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities
 {
-    class AOSClient
+        class AOSClient : IAOSClient
```
karimelazzouni, line 13: Extra space and include the access modifier here

src/CLI/DeleteWorkloads.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using ScaleUnitManagement.WorkloadSetupOrchestrator;
+using System;
+using CLIFramework;
+using ScaleUnitManagement.Utilities;
+
+namespace CLI
+{
+    class DeleteWorkloads
```
karimelazzouni, line 10: Explicit access modifier here

src/CLI/MoveWorkloads.cs
```diff
@@ -0,0 +1,47 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using ScaleUnitManagement.Utilities;
+using ScaleUnitManagement.WorkloadSetupOrchestrator;
+using System;
+
+namespace CLI
+{
+    class MoveWorkloads
```
karimelazzouni, line 9: Ditto

src/CLI/WorkloadMovementStatus.cs
```diff
@@ -0,0 +1,39 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CLIFramework;
+using ScaleUnitManagement.Utilities;
+using ScaleUnitManagement.WorkloadSetupOrchestrator;
+
+namespace CLI
+{
+    class WorkloadMovementStatus
```
karimelazzouni, line 9: Ditto

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -26,6 +26,13 @@ private static CloudAndEdgeConfiguration UserConfiguration()
             return UserConfig;
         }
 
+        internal static Func<CloudAndEdgeConfiguration> UserConfigImplementation { get; set; } = DefaultUserConfiguration;
+
+        private static CloudAndEdgeConfiguration UserConfiguration()
```
karimelazzouni, line 31: nit: `GetUserConfiguration()`

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -15,7 +15,7 @@ public class Config
 
         private static CloudAndEdgeConfiguration UserConfig { get; set; }
 
-        private static CloudAndEdgeConfiguration UserConfiguration()
+        private static CloudAndEdgeConfiguration DefaultUserConfiguration()
```
karimelazzouni, line 18: nit: `GetDefaultUserConfiguration()`

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -26,6 +26,13 @@ private static CloudAndEdgeConfiguration UserConfiguration()
             return UserConfig;
         }
 
+        internal static Func<CloudAndEdgeConfiguration> UserConfigImplementation { get; set; } = DefaultUserConfiguration;
```
karimelazzouni, line 29: Not entirely sure why we're keeping this property and not just `return GetDefaultUserConfiguration()` below?
I understand that it gives more flexibility in returning different user configurations dynamically (by potentially setting the `UserConfigImplementation` property to another function), but it seems like it's never set anywhere aside from the unit tests below.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/AOSClient.cs
```diff
@@ -111,6 +111,35 @@ public async Task<List<WorkloadInstance>> WriteWorkloadInstances(List<WorkloadIn
             }
         }
 
+        public async Task<List<WorkloadInstance>> DeleteWorkloadInstances(List<WorkloadInstance> workloadInstances)
+        {
+            var msg = new HttpRequestMessage(HttpMethod.Post, $"{requestPathPrefix}/api/services/SysWorkloadServices/SysWorkloadInstanceService/delete/");
+
+            // Double serialize the payload in order to avoid serialization complication issues on the AOS.
```
karimelazzouni, line 118: Is the term `AOS` public knowledge? Or is it internal jargon? 
If the latter is the case, maybe replace this with 

`...to avoid serialization complication issues on the application side`
Or
`...to avoid serialization complication issues on FinOps side`

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/AOSClient.cs
```diff
@@ -111,6 +111,35 @@ public async Task<List<WorkloadInstance>> WriteWorkloadInstances(List<WorkloadIn
...
+            var serializedWlInstances = Newtonsoft.Json.JsonConvert.SerializeObject(workloadInstances);
+            var serializedAsString = Newtonsoft.Json.JsonConvert.SerializeObject(serializedWlInstances);
+
+            // Wrap in object that allows the AOS to map in to method params on the service class.
+            // runSync is set to false because we don't want to run CDX infrastructure setup when deleting workload instances.
```
karimelazzouni, line 123: IMO, a lot of the terms/details in this comment are irrelevant to the external audience. 
I suggest removing/amending them.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/MovementState.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+
+namespace ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities
+{
+    class MovementState
```
karimelazzouni, line 5: Ditto

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/MovementState.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private readonly string state;
+        private readonly string workloadMovementCompletedState = "ScaleUnitWorkloadMovementCompletedState";
+
+        public MovementState(string state)
+        {
+            this.state = state;
+        }
+
+        public string getStatus()
```
karimelazzouni, line 15: Pascal case. Should be: `GetStatus()`

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/IAOSClient.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CloudAndEdgeLibs.AOS;
+using CloudAndEdgeLibs.Contracts;
+
+namespace ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities
+{
+    public interface IAOSClient
```
karimelazzouni, line 9: Why not reuse the `IAOSClient` in `CloudAndEdgeLibs`? Is there a reason why we want to duplicate it here?

src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
```diff
@@ -0,0 +1,101 @@
...
+                int count = 0;
+                foreach (MovementState state in stateList)
+                {
+                    Console.WriteLine($"{workloadInstanceIdWithNameList[count].Name} Id : {workloadInstanceIdWithNameList[count].WorkloadInstanceId} Workload installation status: {state.getStatus()}");
+                    count++;
+                }
+            }, "Movement status");
+        }
+
+        private async Task<string> getMovementState(AOSClient aosClient, WorkloadInstance workloadInstance)
```
karimelazzouni, line 86: Pascal case. Should be `GetMovementStatus(...)`

src/ScaleUnitManagementTests/ConfigurationHelper.cs
```diff
@@ -0,0 +1,53 @@
+
+using System;
+using System.Collections.Generic;
+using ScaleUnitManagement.Utilities;
+
+namespace ScaleUnitManagementTests
+{
+    class ConfigurationHelper
```
karimelazzouni, line 8: Ditto

karimelazzouni (changes requested):

src/ScaleUnitManagement/Utilities/Config.cs
Antonstockmarr: This part is only added to allow tests to set their own configurations. Without it, I had trouble running tests on methods that make use of the configuration file, and Andreas suggested this pattern as the best way to solve that.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
```diff
@@ -0,0 +1,101 @@
...
+                ScaleUnitInstance scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
+                AOSClient aosClient = await AOSClient.Construct(scaleUnit);
+                List<WorkloadInstance> workloadInstances = await aosClient.GetWorkloadInstances();
+                List<MovementState> stateList = new List<MovementState>();
+                List<WorkloadInstanceIdWithName> workloadInstanceIdWithNameList = Config.WorkloadInstanceIdWithNameList();
+
+                foreach (WorkloadInstance workloadInstance in workloadInstances)
+                {
+                    string state = await getMovementState(aosClient, workloadInstance);
+                    stateList.Add(new MovementState(state));
```
AndreasHassing, line 74: MovementState should also contain workload instance name (and/or ID) so you don't need to iterate the named list.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
AndreasHassing, line 86: Java! ☕

src/ScaleUnitManagementTests/ConfigurationHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+
+        public CloudAndEdgeConfiguration ConstructTestConfiguration()
+        {
+            ScaleUnitInstance spoke = new ScaleUnitInstance()
```
AndreasHassing, line 13: Should use `var`

src/ScaleUnitManagementTests/ConfigurationHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+            };
+
+            ScaleUnitInstance hub = new ScaleUnitInstance()
+            {
+                ScaleUnitId = "@@"
+            };
+
+            List<ScaleUnitInstance> scaleUnitInstances = new List<ScaleUnitInstance>();
+            scaleUnitInstances.Add(spoke);
+            scaleUnitInstances.Add(hub);
```
AndreasHassing, line 25: Could use collection initializer syntax:
```csharp
var scaleUnitInstances = new List<..>
{
  spoke,
  hub,
};
```

src/ScaleUnitManagementTests/ScaleUnitManagementTests.csproj
```diff
@@ -6,8 +6,16 @@
...
 
+  <ItemGroup>
+    <Compile Remove="FakesAssemblies\**" />
+    <EmbeddedResource Remove="FakesAssemblies\**" />
+    <None Remove="FakesAssemblies\**" />
+  </ItemGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Include="Microsoft.QualityTools.Testing.Fakes" Version="16.7.4-beta.20330.2" />
```
AndreasHassing, line 17: Consider removing as you are not using shims

AndreasHassing: And other artifacts

src/ScaleUnitManagementTests/WorkloadDeletionTest.cs
```diff
@@ -0,0 +1,74 @@
...
+            WorkloadInstance workloadInstance = new WorkloadInstance { Id = configuredWorkload.WorkloadInstanceId };
+            workloadInstance.ExecutingEnvironment.Add(new TemporalAssignment
+            {
+                EffectiveDate = DateTime.UtcNow,
+                Environment = new PhysicalEnvironmentReference() { ScaleUnitId = scaleUnitId },
+            });
+            toBeReturnedWorkloadInstances.Add(workloadInstance);
+
+            aosClient.Setup(x => x.GetWorkloadInstances())
+                .Returns(() => Task.FromResult(new List<WorkloadInstance>(toBeReturnedWorkloadInstances)));
```
AndreasHassing, line 52: I think there is a ReturnsAsync to avoid Task.FromResult

AndreasHassing (changes requested):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
karimelazzouni: 😂

src/ScaleUnitManagement/Utilities/Config.cs
karimelazzouni: Another known pattern is to create a configuration provider that is responsible for passing the needed configuration into this class (`IConfigurationProvider`). 

Then, you can pass a mocked/test implementation of this interface in your tests to use whichever values you want. Either ways, both patterns achieve the same result 😊

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/IAOSClient.cs
karimelazzouni: Discussed offline. In order not to expose additional functionality that's irrelevant to this repo, we're going to keep the duplicate interface.

src/ScaleUnitManagement/Utilities/Config.cs
Antonstockmarr: Yes, that could also be a solution. However, I think that would also require some changes to the config class, since the mocked implementation will have to get around the Validation method.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadDeleter.cs
```diff
@@ -0,0 +1,61 @@
...
+                {
+                    Console.WriteLine($"No workloads to delete on scale unit {scaleUnit.ScaleUnitId}");
+                }
+                else
+                {
+                    List<WorkloadInstanceIdWithName> workloadInstanceIdWithNameList = Config.WorkloadInstanceIdWithNameList();
+
+                    foreach (WorkloadInstance workloadInstance in workloadInstances)
+                    {
+                        WorkloadInstanceIdWithName workload = workloadInstanceIdWithNameList.Find(wl => wl.WorkloadInstanceId == workloadInstance.Id);
```
karimelazzouni, line 50: nit: 
```suggestion
                        WorkloadInstanceIdWithName workload = workloadInstanceIdWithNameList.Find(wl => wl.WorkloadInstanceId.Equals(workloadInstance.Id));
```

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/AOSClient.cs
AndreasHassing: AOS is public knowledge for external developers as well: https://docs.microsoft.com/en-us/dynamicsax-2012/developer/aos-overview

AndreasHassing (changes requested):
There are still a few comments from the first review that were not resolved 😊.

src/CLI/MoveWorkloads.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        public static async Task MoveAllWorkloads(int input, string selectionHistory)
```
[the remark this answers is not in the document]
AndreasHassing: It's now in a sub-menu so no longer needed 😊.

src/CLI/Program.cs
```diff
@@ -12,11 +12,12 @@ public static async Task Main(string[] args)
             CheckForAdminAccess.ValidateCurrentUserIsProcessAdmin();
 
             var enableScaleUnitFeatureOption = new CLIOption() { Name = "Initialize the hybrid topology", Command = EnableScaleUnitFeature.SelectScaleUnit };
-            var installationMenuOption = new CLIOption() { Name = "Show installation options", Command = InstallationMenu.Show };
-            var movementMenuOption = new CLIOption() { Name = "Show movement options", Command = MovementMenu.Show };
+            var configureEnvironmentOption = new CLIOption() { Name = "Prepare environments for workload installation", Command = ConfigureEnvironment.Show };
+            var installationMenuOption = new CLIOption() { Name = "Show installation options", Command = WorkloadInstallationMenu.Show };
```
AndreasHassing, line 16: "Show workload installation options", and below: "Show workload movement options"

AndreasHassing (approved):

Issue #10, opened 2020-12-04
Replace uses of PowerShell scripts with C#-native code


AndreasHassing:
This should be done to improve exception/error-handling and will also improve code readability as we won't need the special `Quotes` injected into strings passed to PowerShell.

- [x] DbSync and ConfigEncryptor should be invoked via [`Process.Start`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.start?view=net-5.0).
- [x] `Invoke-SqlCmd` should be replaced with use of [`SqlCommand`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlcommand?view=sqlclient-dotnet-core-2.1) et al.
- [x] `(Start|Stop)-Service` should be replaced with use of [`ServiceController`](https://docs.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicecontroller).
- [x] Setting IIS bindings with `Set-ItemProperty` should be replaced with use of [`ServerManager`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.administration.servermanager). Fixed in #17.

AndreasHassing:
Find usages here: https://github.com/microsoft/SCMScaleUnitDevTools/search?q=%22new+CommandExecutor%22

PR #70, opened 2021-08-13
Replace use of Powershell scripts using SqlClient and Process.Start

Antonstockmarr:
CommandExecutor can now execute commands without PowerShell using `Process.Start`.
`Invoke-SqlCmd` has been removed. Instead `SqlCommand` is used to query the database.

Resolves remaining items on #10

### DIFF
--- a/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/RunDBSync.cs
+++ b/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/RunDBSync.cs
@@ -27,23 +27,23 @@ namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
 
                 string dbSyncTool = Path.Combine(scaleUnit.ServiceVolume, @"AOSService\PackagesLocalDirectory\bin\syncengine.exe");
                 string metaBinariesPath = Path.Combine(scaleUnit.ServiceVolume, @"AOSService\PackagesLocalDirectory");
+                string outputFile = $"{scaleUnit.AxDbName}_DbSync.log";
 
-                string cmd =
-                    dbSyncTool
-                    + " -syncmode=" + CommandExecutor.Quotes + "fullall" + CommandExecutor.Quotes
-                    + " -metadatabinaries=" + CommandExecutor.Quotes + metaBinariesPath + CommandExecutor.Quotes
-                    + " -connect=" + CommandExecutor.Quotes + $"Data Source=localhost;Initial Catalog={scaleUnit.AxDbName};Integrated Security=True;Enlist=True;Application Name=AXVSSDK" + CommandExecutor.Quotes
-                    + " -verbosity=" + CommandExecutor.Quotes + "Diagnostic" + CommandExecutor.Quotes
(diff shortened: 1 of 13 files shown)


AndreasHassing (changes requested):
RFC, looks good other than a few nits 😊.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/RunDBSync.cs
```diff
@@ -27,23 +27,23 @@ public void Run()
...
-                    dbSyncTool
-                    + " -syncmode=" + CommandExecutor.Quotes + "fullall" + CommandExecutor.Quotes
-                    + " -metadatabinaries=" + CommandExecutor.Quotes + metaBinariesPath + CommandExecutor.Quotes
-                    + " -connect=" + CommandExecutor.Quotes + $"Data Source=localhost;Initial Catalog={scaleUnit.AxDbName};Integrated Security=True;Enlist=True;Application Name=AXVSSDK" + CommandExecutor.Quotes
-                    + " -verbosity=" + CommandExecutor.Quotes + "Diagnostic" + CommandExecutor.Quotes
-                    + " -scaleUnitOptionsAsJson=" + CommandExecutor.Quotes + "{" + CommandExecutor.Quotes + "IsScaleUnitFeatureEnabled" + CommandExecutor.Quotes + ": true, " + CommandExecutor.Quotes + "scaleUnitMnemonics" + CommandExecutor.Quotes + ":" + $"'{scaleUnit.ScaleUnitId}'" + " }" + CommandExecutor.Quotes
-                    + " -triggerOptionsAsJson=" + CommandExecutor.Quotes + "{" + CommandExecutor.Quotes + "IsEnabled" + CommandExecutor.Quotes + ": true}" + CommandExecutor.Quotes
-                    + $" > {scaleUnit.AxDbName}_DbSync.log";
+                string args =
+                    "-syncmode=\"fullall\""
```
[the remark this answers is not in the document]
Antonstockmarr: I did test this, and it does work. I believe the reason is that before the operation was called using Process.Start(), with file name set to "powershell". This meant that the string had to be parsed first by the Process object, and then afterwards by PowerShell. So the qoutes needed to be wrapped twice. But now the command is run directly by Process.Start(), with the file name set directly to the dbsSync executable. So the double wrapped quotes are not needed.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/IISAdministrationHelper.cs
```diff
@@ -133,7 +133,7 @@ private static void CloneScaleUnitWebRoot(ScaleUnitInstance scaleUnit)
                 $@"robocopy {Config.HubScaleUnit().SiteRoot()} {scaleUnit.SiteRoot()} /MIR /MT /log+:{robocopyLogPath};";
 
             CommandExecutor ce = new CommandExecutor(cmd);
-            ce.RunCommand(new List<int>() { 0, 1 });
+            ce.RunCommand(new int[] { 0, 1 });
```
AndreasHassing, line 136: The cool thing about `params` is that you can now use `RunCommand` like:

```csharp
ce.RunCommand(); // assumes an empty array
ce.RunCommand(1); // passes in a: `new int[] { 1 }`
ce.RunCommand(0, 1); // passes in a: `new int[] { 0, 1 }`
```

AndreasHassing (changes requested):
Fix use of `params` and we're golden. Thanks for this Anton! 👍

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/IISAdministrationHelper.cs
Antonstockmarr: Ahh, okay that's really cool!

AndreasHassing (approved):

PR #72, opened 2021-08-16
Resets the tool after each action

Antonstockmarr:
The tool will start over after performing an action. An "Exit" option has been added to the root menu.
#patch

### DIFF
--- a/src/CLIFramework/CLIMenu.cs
+++ b/src/CLIFramework/CLIMenu.cs
@@ -115,5 +115,12 @@ namespace CLIFramework
 
             return String.IsNullOrEmpty(input) || input == "y" || input == "yes";
         }
+
+        public static void PressToContinue()
+        {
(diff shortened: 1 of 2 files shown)


AndreasHassing (approved):

src/CLIFramework/CLIMenu.cs
```diff
@@ -115,5 +115,12 @@ public static bool YesNoPrompt(string message)
 
             return String.IsNullOrEmpty(input) || input == "y" || input == "yes";
         }
+
+        public static void PressToContinue()
```
AndreasHassing, line 119: Where is this used_

AndreasHassing: Never mind, I see it.

AndreasHassing (approved):

PR #78, opened 2021-08-30
Disable scale unit feature

Omaaarz:
Added a capability to disable scale unit feature on scale units and remove triggers

### DIFF
--- a/src/CLI/DisableScaleUnitFeature.cs
+++ b/src/CLI/DisableScaleUnitFeature.cs
@@ -1,0 +1,46 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CLIFramework;
+using ScaleUnitManagement.Utilities;
+using System;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Common;
+
+namespace CLI
+{
+    internal class DisableScaleUnitFeature
+    {
+        private static List<ScaleUnitInstance> sortedScaleUnits;
+        public static async Task Show(int input, string selectionHistory)
+        {
+            var options = new List<CLIOption>();
+
+            sortedScaleUnits = Config.ScaleUnitInstances();
+            sortedScaleUnits.Sort();
+
+            foreach (ScaleUnitInstance scaleUnit in sortedScaleUnits)
+            {
+                options.Add(new CLIOption() { Name = scaleUnit.PrintableName(), Command = RunDisableScaleUnitFeature });
+            }
+
+            var screen = new CLIScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to disable scale unit feature on?: ");
+            await CLIMenu.ShowScreen(screen);
+        }
+
+        private static Task RunDisableScaleUnitFeature(int input, string selectionHistory)
+        {
+            using (var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId))
+            {
+                try
+                {
+                    new RunDBSync().Run(false);
+                }
+                catch (Exception ex)
+                {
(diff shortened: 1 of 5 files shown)


src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/RunDBSync.cs
```diff
@@ -17,7 +17,7 @@ public float Priority()
             return 3F;
         }
 
-        public void Run()
+        public void Run(bool isScaleUnitFeatureEnabled = true)
         {
             Console.WriteLine("Executing DbSync");
```
Antonstockmarr, line 22: Do we still want to print this if we are disabling scale unit feature?

Antonstockmarr (approved):

src/CLI/SyncDB.cs
```diff
@@ -26,20 +26,21 @@ public static async Task Show(int input, string selectionHistory)
             await CLIMenu.ShowScreen(screen);
         }
 
-        private static async Task RunSyncDB(int input, string selectionHistory)
+        private static Task RunSyncDB(int input, string selectionHistory)
         {
-            ScaleUnitDBSync scaleUnitDBSync = new ScaleUnitDBSync();
```
AndreasHassing, line 31: Is the `ScaleUnitDBSync` class no longer needed? If so, can probably remove in this PR.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/SharedWebConfig.cs
```diff
@@ -16,8 +16,8 @@ public static void Configure(WebConfig webConfig)
                     webConfig.UpdateXElementIfExists("AzureStorage.StorageConnectionString", scaleUnit.AzureStorageConnectionString);
             }
 
-            webConfig.AddKey("ScaleUnit.InstanceID", scaleUnit.ScaleUnitId);
-            webConfig.AddKey("ScaleUnit.Enabled", "true");
+            webConfig.AddKey("ScaleUnit.InstanceID", isScaleUnitFeatureEnabled ? scaleUnit.ScaleUnitId : "");
```
AndreasHassing, line 19: Does it matter what this value is if `ScaleUnit.Enabled == false`?

AndreasHassing (approved):
Minor comments 😊. Approving regardless.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/SharedWebConfig.cs
Omaaarz: It shouldn't but wanted to play it safe if there are any places where we miss checking the enablement of the feature before using the mnemonic.

src/CLI/SyncDB.cs
Omaaarz: The class is still being used as part of EnableScaleUnitFeature option. It's a wrapper that we don't need in our case here.

src/CLI/DisableScaleUnitFeature.cs
```diff
@@ -0,0 +1,53 @@
...
+        private static Task RunDisableScaleUnitFeature(int input, string selectionHistory)
+        {
+            using (var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId))
+            {
+                try
+                {
+                    new StopServices().Run();
+                    using (var webConfig = new WebConfig())
+                    {
+                        SharedWebConfig.Configure(webConfig, false);
```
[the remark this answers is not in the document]
Omaaarz: Makes sense.

PR #83, opened 2021-09-01
Update web config file with new scale unit id based on user config changes

Omaaarz:
Update web config file with new scale unit id based on user config changes

### DIFF
--- a/src/CLI/SetupToolsOptions/UpdateScaleUnitId.cs
+++ b/src/CLI/SetupToolsOptions/UpdateScaleUnitId.cs
@@ -23,7 +23,7 @@ namespace CLI.SetupToolsOptions
                 options.Add(new CLIOption() { Name = scaleUnit.PrintableName(), Command = RunUpdateScaleunitId });
             }
 
-            var screen = new CLIScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to update its scale unit id?: ");
+            var screen = new CLIScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to update the scale unit id of?: ");
             await CLIMenu.ShowScreen(screen);
         }
 
--- a/src/CLI/SetupToolsOptions/UpdateScaleUnitId.cs
+++ b/src/CLI/SetupToolsOptions/UpdateScaleUnitId.cs
@@ -1,0 +1,52 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CLIFramework;
+using ScaleUnitManagement.Utilities;
+using System;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Common;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
+
+namespace CLI.SetupToolsOptions
+{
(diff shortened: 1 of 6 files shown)


src/CLI/SetupToolsOptions/UpdateScaleUnitId.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            sortedScaleUnits = Config.NonHubScaleUnitInstances();
+            sortedScaleUnits.Sort();
+
+            foreach (ScaleUnitInstance scaleUnit in sortedScaleUnits)
+            {
+                options.Add(new CLIOption() { Name = scaleUnit.PrintableName(), Command = RunUpdateScaleunitId });
+            }
+
+            var screen = new CLIScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to update its scale unit id?: ");
```
AndreasHassing, line 26:
```suggestion
            var screen = new CLIScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to update the scale unit id of?: ");
```

AndreasHassing (approved):
Minor comment, LGTM.

AndreasHassing:
Is this for restore? Does DbSync need to run after changing scale unit ID? Not sure I understand exactly why or when this feature is needed.

Omaaarz:
> Is this for restore? Does DbSync need to run after changing scale unit ID? Not sure I understand exactly why or when this feature is needed.

@AndreasHassing Yes, this is for restore. This step will be done before even importing the DB and that's why we shouldn't be doing DBsync at this step. Please see [here](https://microsoft-my.sharepoint.com/:p:/p/omeissaa/EY2hBbLmMipGr2rVVqjlj6YB6ZSYKfBqkHEWQc6AY6FA4A?e=bLuHy1) the scenario for UAT to Devbox(Both CnE).

Importing the DB should be followed by DBsync. 

Please let me know if you have any concerns.

AndreasHassing:
> > Is this for restore? Does DbSync need to run after changing scale unit ID? Not sure I understand exactly why or when this feature is needed.
> 
> @AndreasHassing Yes, this is for restore. This step will be done before even importing the DB and that's why we shouldn't be doing DBsync at this step.

I'm wondering if it makes sense to have this action from the tool stand-alone, or if it should be an integrated part of restore? You may have a design in mind that puts my concerns to rest, if so feel free to merge this 😊.

Omaaarz:
> > > Is this for restore? Does DbSync need to run after changing scale unit ID? Not sure I understand exactly why or when this feature is needed.
> > 
> > 
> > @AndreasHassing Yes, this is for restore. This step will be done before even importing the DB and that's why we shouldn't be doing DBsync at this step.
> 
> @Omaaarz: I'm wondering if it makes sense to have this action from the tool stand-alone, or if it should be an integrated part of restore? You may have a design in mind that puts my concerns to rest, if so feel free to merge this 😊.

I'm trying to have the experience here as close as possible to steps needed in SUM for (Prod -> UAT) operations. That's why I'm keeping it as a standalone step now.

PR #84, opened 2021-09-01
Make seperate API write request for every workload

### LINKED ISSUE #82: WriteWorkloadInstances should always be called on one workload at a time
Right now we call WriteWorkloadInstances on a list of all the workloads at the same time. This can very quickly become a problem, as there is a limit to how much data can be processed in a single request.

This should be changed here
https://github.com/microsoft/SCMScaleUnitDevTools/blob/5c96b97ea6cd31add2d1dc70d1bbe4bdfbf0d381/src/ScaleUnitManagement/WorkloadSetupOrchestrator/HubWorkloadInstaller.cs#L54
and here
https://github.com/microsoft/SCMScaleUnitDevTools/blob/19a76e30588d6478117e6a45eb6db68f93fde5d8/src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs#L34

Antonstockmarr:
Asynchronously make an API request for every workloadInstance when writing workloads, instead of writing all of them at the same time. Manually tested in the SingleOneBox environment.
closes #82 
#patch

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
@@ -31,7 +31,7 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator
                     }
                     workloadInstance.ExecutingEnvironment.Add(CreateTemporalAssignment(moveToId, movementDateTime));
                 }
-                await aosClient.WriteWorkloadInstances(workloadInstances);
+                await Task.WhenAll(workloadInstances.Select(workload => aosClient.WriteWorkloadInstances(new List<WorkloadInstance> { workload })));
             }, $"Move workloads from scaleUnit {scaleUnit.ScaleUnitId} to hub");
         }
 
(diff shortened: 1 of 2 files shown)


src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
```diff
@@ -31,7 +31,7 @@ public async Task MoveWorkloads(string moveToId, DateTime movementDateTime)
                     }
                     workloadInstance.ExecutingEnvironment.Add(CreateTemporalAssignment(moveToId, movementDateTime));
                 }
-                await aosClient.WriteWorkloadInstances(workloadInstances);
+                await Task.WhenAll(workloadInstances.Select(workload => aosClient.WriteWorkloadInstances(new List<WorkloadInstance> { workload })));
```
AndreasHassing, line 34: Consider splitting into a loop. Not sure how AX will handle the parallelism going on here and it's safer to just do it synchronously one-by-one. This also gives better error reporting as we'll know by the point in time of the loop which one is causing problems.

AndreasHassing (changes requested):

AndreasHassing (approved):
LGTM :shipit:.

Can you put what kinds of validations you did in the PR description, for visibility of due diligence.

PR #85, opened 2021-09-06
Validate the AAD apps used in the configuration

### LINKED ISSUE #79: Validate InterAOS AAD app and tool AAD app are not the same
In #76 the same AAD app was used for both the developer tool and inter-AOS communication, this should be enforced by the tool to avoid it from happening. The documentation dictates that two apps should be used for these different purposes.

The error went something like (from a request made from Spoke to Hub):
`An exception occured when authorizing - Entry point security access denied to current user for operation 'getInitializationData' on service 'ScaleUnitInitializationService'.`

Antonstockmarr:
The AAD app for a scale unit must never be the same as the AAD app for the inter-AOS communications.
#patch
Resolves #79


src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -157,6 +157,12 @@ void ValidateValue(string fieldName, string fieldValue)
                 ValidateValue("AppSecret", scaleUnit.AuthConfiguration.AppSecret);
                 ValidateValue("Authority", scaleUnit.AuthConfiguration.Authority);
 
+                if (scaleUnit.AuthConfiguration.AppId.Equals(InterAOSAppId()))
```
Omaaarz, line 160:
```suggestion
                if (scaleUnit.AuthConfiguration.AppId.Equals(InterAOSAppId(), StringComparison.OrdinalIgnoreCase))
```

src/ScaleUnitManagement/Utilities/Config.cs
```diff
@@ -157,6 +157,12 @@ void ValidateValue(string fieldName, string fieldValue)
                 ValidateValue("AppSecret", scaleUnit.AuthConfiguration.AppSecret);
                 ValidateValue("Authority", scaleUnit.AuthConfiguration.Authority);
 
+                if (scaleUnit.AuthConfiguration.AppId.Equals(InterAOSAppId(), StringComparison.OrdinalIgnoreCase))
+                {
+                    hasAnyError = true;
+                    Console.Error.WriteLine("The scale unit with id \"" + (scaleUnit.ScaleUnitId) + "\" is running with the same AAD app as the interAOS AAD app. The interAOS communication AAD app should always be different what is used for the scale units.");
```
AndreasHassing, line 163:
```suggestion
                    Console.Error.WriteLine($"The scale unit with id ""{scaleUnit.ScaleUnitId}"" is configured to have the same AAD app as the interAOS AAD app. The interAOS communication AAD app should always be different from what is used for the scale units.");
```

AndreasHassing (approved):
Approving with suggestion.

src/ScaleUnitManagement/Utilities/Config.cs
AndreasHassing: Wups, it's `$@"..."`, otherwise the `""` double quote won't work :).

Omaaarz (approved):

PR #86, opened 2021-09-08
Add capability to import workloads from different Azure Storage

Antonstockmarr:
Using a SAS token, the user can copy an Azure blob into one of the scale unit containers.

Tested manually in the singleOneBox environment by importing workloads between hub and spoke, with SAS tokens created on the Azure Storage page.

### DIFF
--- a/src/CLI/SetupToolsOptions/CleanUpStorageAccount.cs
+++ b/src/CLI/SetupToolsOptions/CleanUpStorageAccount.cs
@@ -30,8 +30,8 @@ namespace CLI.SetupToolsOptions
         {
             using (var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId))
             {
-                var accountCleaner = new StorageAccountManager();
-                await accountCleaner.CleanStorageAccount();
+                var storageAccountManager = new StorageAccountManager();
+                await storageAccountManager.CleanStorageAccount();
             }
 
             Console.WriteLine("Done");
--- a/src/CLI/SetupToolsOptions/CleanUpStorageAccount.cs
+++ b/src/CLI/SetupToolsOptions/CleanUpStorageAccount.cs
@@ -30,7 +30,7 @@ namespace CLI.SetupToolsOptions
         {
             using (var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId))
             {
-                var accountCleaner = new StorageAccountCleaner();
+                var accountCleaner = new StorageAccountManager();
                 await accountCleaner.CleanStorageAccount();
             }
 
--- a/src/CLI/SetupToolsOptions/ImportWorkloadBlob.cs
+++ b/src/CLI/SetupToolsOptions/ImportWorkloadBlob.cs
@@ -1,0 +1,48 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CLIFramework;
+using ScaleUnitManagement.Utilities;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Common;
+using System;
+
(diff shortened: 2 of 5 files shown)


src/CLI/SetupToolsOptions/CleanUpStorageAccount.cs
```diff
@@ -30,7 +30,7 @@ private static async Task CleanUpScaleUnitStorageAccount(int input, string selec
         {
             using (var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId))
             {
-                var accountCleaner = new StorageAccountCleaner();
+                var accountCleaner = new StorageAccountManager();
```
Omaaarz, line 33: Name of variable needs to be adjusted to match the new name of the class

src/CLI/SetupToolsOptions/ImportWorkloadBlob.cs
```diff
@@ -0,0 +1,48 @@
...
+            var screen = new CLIScreen(options, selectionHistory, "Please select the scale unit you would like to import workloads to:\n", "\nScale unit storage to import to: ");
+            await CLIMenu.ShowScreen(screen);
+        }
+
+        private static async Task ImportWorkloadBlobFromSasToken(int input, string selectionHistory)
+        {
+            var sasToken = CLIMenu.EnterValuePrompt("Please paste in the blob SAS URL for the blob storage that the workloads should be copied from:");
+            using (var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId))
+            {
+                var accountCleaner = new StorageAccountManager();
```
Omaaarz, line 34:
```suggestion
                var accountManager = new StorageAccountManager();
```

src/ScaleUnitManagement/WorkloadSetupOrchestrator/StorageAccountManager.cs
```diff
@@ -1,27 +1,60 @@
...
+                sasUri = new Uri(sasUrlString);
+            }
+            catch (Exception)
+            {
+                throw new Exception($"{sasUrlString} is not a valid Uri");
+            }
+
+            BlobClientOptions options = null;
+            var sourceBlobClient = new BlobClient(sasUri, options);
+            var blobContainerName = sourceBlobClient.BlobContainerName;
```
Omaaarz, line 44: We shouldn't get the name of blob file or container from the given SAS. We know the target container name and blob name (You can get them from your storage account). The given SAS might have a different blob name and/or different container name so we shouldn't rely on them

Omaaarz (changes requested):

src/CLI/SetupToolsOptions/ImportWorkloadBlob.cs
```diff
@@ -0,0 +1,48 @@
...
+                options.Add(new CLIOption() { Name = scaleUnit.PrintableName(), Command = ImportWorkloadBlobFromSasToken });
+            }
+
+            var screen = new CLIScreen(options, selectionHistory, "Please select the scale unit you would like to import workloads to:\n", "\nScale unit storage to import to: ");
+            await CLIMenu.ShowScreen(screen);
+        }
+
+        private static async Task ImportWorkloadBlobFromSasToken(int input, string selectionHistory)
+        {
+            var sasToken = CLIMenu.EnterValuePrompt("Please paste in the blob SAS URL for the blob storage that the workloads should be copied from:");
```
Omaaarz, line 31: NIT
```suggestion
            var sasToken = CLIMenu.EnterValuePrompt("Please paste in the blob SAS URL for the blob that the workloads should be copied from:");
```

Omaaarz (approved):
LGTM. left minor comment

PR #89, opened 2021-09-09
Refactoring

Antonstockmarr:
Refactor the cli library, making it easier to maintain and more organized. All menus where the user must give an input are now instances of the abstract class DevToolMenu.

Also fix a large amount of small Nits all around the project, mostly trivial changes, like changing explicit variable into "var"

Verified by manually preparing and installing workloads on the singleOneBox environment.

#patch


src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/StepFactory.cs
```diff
@@ -16,12 +16,12 @@ private IEnumerable<Type> FindClassesDeriving(Type superType)
 
         public List<IStep> GetStepsOfType<T>()
         {
-            var steps = new List<IStep>();
+            var objects = new List<IStep>();
```
Omaaarz, line 19: Is this intentional? if yes, why?

src/CLI/DevToolMenu.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            return GetSortedScaleUnits()[index].ScaleUnitId;
+        }
+
+        protected CLIOption Option(string name, Func<int, string, Task> command)
+        {
+            return new CLIOption { Name = name, Command = command };
+        }
+
+        protected List<CLIOption> SelectScaleUnitOptions(Func<int, string, Task> command)
```
Omaaarz, line 35: I'm thinking if we could have an optional parameter here to pass scale unit instances that this menu works on. That would help if we want to specify specific scale units that are valid for taking action instead of dealing with such type of devtools differently in their Show method. What do you think?

Omaaarz (changes requested):
Looks good. I'm suggesting one improvement that we can have to help accommodating actions that would be handling not all scale units but subset of them

src/ScaleUnitManagementTests/WorkloadInstanceTopologicalSortUtilTest.cs
```diff
@@ -10,21 +10,21 @@ namespace ScaleUnitManagementTests
     [TestClass]
     public class WorkloadInstanceTopologicalSortUtilTest
     {
-        private readonly string SYS = "Sys";
-        private readonly string MES = "Mes";
-        private readonly string WES = "Wes";
+        private readonly string sys = "Sys";
```
AndreasHassing, line 13: These should be constants and capital (that's how they're usually represented).

AndreasHassing (changes requested):
You're not going to like this 😅, but; unless you agreed with Omar to change a lot of variable declarations from `T` to `var`, you should revert those changes.

The only time to use `var` is when the type is immediately obvious from the right-hand side - which it is when:
* you're `new`'ing up an object:
   ```csharp
   var x = new X();
   ```
* you're type-casting:
   ```csharp
   var y = (Y)yThing;
   ```

Or in rare cases where the type is fairly obvious and typing out the type would yield sub-par readability (cases of async enumeration and LINQ expression trees come to mind).

In C#9 you can do:
```csharp
Dictionary<string, string> d = new()
{
    ["Hey!"] = "Stop waving..",
};
```
Further reducing the value of (in my opinion) the `var` keyword.

To be consistent in a .NET code base, you can consider loosely following [the C# coding conventions](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#implicitly-typed-local-variables) (not everyone does, so stay relaxed about it).

src/CLI/DevToolMenu.cs
AndreasHassing: Or, its signature could be:
```csharp
// takes care of creating a `ScaleUnitContext`:
protected List<CLIOption> SelectScaleUnitOptions(Func<Task> command);

// child-class handles creating a `ScaleUnitContext`, but is passed the scale unit:
protected List<CLIOption> SelectScaleUnitOptions(Func<ScaleUnitInstance, Task> command);
```

AndreasHassing:
@Omaaarz can you pitch in with your thoughts on the `T => var` comment I made? If you're OK with having `var` as a default, I'm fine with it too - your call.

src/ScaleUnitManagementTests/WorkloadInstanceTopologicalSortUtilTest.cs
AndreasHassing: In fact, both the string value and the field names.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/StepFactory.cs
Antonstockmarr: It started out with the name objects, then I changed it to steps because I thought it was more intuitive without the explicit type. But then I changed my mind, because this code is used to instantiate the step objects.

Antonstockmarr:
The reason why I changed the types to being implicit is that it is treated as a "formatting error" by VS Code, which means that other errors drown in 350 other messages. But I see your point, and VS code can probably be configured to not account for this if we decide to keep the explicit types.

Omaaarz:
> @Omaaarz can you pitch in with your thoughts on the `T => var` comment I made? If you're OK with having `var` as a default, I'm fine with it too - your call.

No I actually agree with you Andreas. When I was looking at the PR yesterday, I was looking at it commit by commit to make it easier for me to review and was planning to commit on the var declaration but forgot after I finished reviewing the other huge commit.

src/CLI/DevToolMenu.cs
Antonstockmarr: I can see the idea of passing the scaleUnits to the function, so it can be used in the case where not all scale units can perform that action. Its also a good idea to let this function construct the scaleUnitContext. But I don't see the benefit of using the other signatures? The purpose of the method is to construct an option for every scale unit, how the command is executed is handled by the CLIController. Can you elaborate on how this would be used, @AndreasHassing ?

AndreasHassing: You're right, I think - just go with whatever feels best 😊. At first glance I thought it would be easy to do, but it seems a bit trickier now that I try to think harder about it.

Omaaarz (approved):

AndreasHassing (approved):

AndreasHassing:
These changes are quite broad throughout the app; consider timeboxing an hour or two into validating almost all available flows in the dev tools before merging this 😊.

Antonstockmarr:
Everything seems to work as expected, so I recommend merging this branch. I do not have the required permission to do so @AndreasHassing / @Omaaarz.

PR #91, opened 2021-09-10
Starting the W3SVC service in case it is stopped

### LINKED ISSUE #62: Start IIS service if stopped
After dbsync, we try to start the AOS IIS sites, but if the w3svc is stopped, it fails. We should start it if it's stopped:

image

Exception message:
```
Microsoft.Web.Administration.ServerManagerException: The World Wide Web Publishing Service (W3SVC) is stopped.
Websites cannot be started unless the World Wide Web Publishing Service (W3SVC) is running.
```

Antonstockmarr:
When starting IIS site AOSServices, the IIS services must be running. This change will start it every time, and if it is already started it will just state that and move on.
#patch
closes #62 

After stopping the services and making sure that W3SVC is not running, this is the output:
image


AndreasHassing:
Can you provide test details? Can you show a screenshot of successful start of `W3SVC`, just to make sure the service name is correct, etc. 😊.

AndreasHassing (approved):

Omaaarz (approved):

PR #98, opened 2021-09-17
Fix cleaning up storage account of LCS-hosted environment, by loading storage string from web.config

### LINKED ISSUE #97: Cannot clear storage account on LCS-hosted devbox
I'm trying to re-enabling the scale unit, when I try to clear the azure storage, I get an error

Scale unit storage to clean up: 2
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'connectionString')
   at Azure.Storage.StorageConnectionString.Parse(String connectionString)
   at Azure.Storage.Blobs.BlobServiceClient..ctor(String connectionString, BlobClientOptions options)
   at ScaleUnitManagement.ScaleUnitFeatureManager.Common.StorageAccountCleaner.DeleteBlobContainers() in /home/runner/work/SCMScaleUnitDevTools/SCMScaleUnitDevTools/src/ScaleUnitManagement/ScaleUnitFeatureManager/Common/StorageAccountCleaner.cs:line 29

Antonstockmarr:
On LCS-hosted environments the Azure storage connection string is not required in the userConfig. In this case, the connection string is loaded from the web.config file. The file is decrypted first.

I do not have access to an LCS-hosted environment, but on my SingleOneBox environment the correct string can be loaded from webconfig, tables are cleared, and the web.config is encrypted afterwards.

#patch
resolves #97


Antonstockmarr:
Can one of you clarify for me:
Can it ever happen that there is an error in the constructor of WebConfig, such that the file is decrypted, but dispose is never invoked? If yes, how can we unsure that web.config can never be left unencrypted?

AndreasHassing:
> @AndreasHassing / @Omaaarz you clarify for me:
> Can it ever happen that there is an error in the constructor of WebConfig, such that the file is decrypted, but dispose is never invoked? If yes, how can we unsure that web.config can never be left unencrypted?

To help ensure that issues won't occur, we need a sequence like:
1. Copy `web.config` to a temporary directory.
2. Decrypt that copy.
3. Make the change to the copy.
4. Re-encrypt it.
5. Copy it back to the `AosService` directory.

src/CLI/SetupToolsOptions/CleanUpStorageAccount.cs
```diff
@@ -28,13 +28,17 @@ public static async Task Show(int input, string selectionHistory)
...
+            try
             {
+                using var context = ScaleUnitContext.CreateContext(sortedScaleUnits[input - 1].ScaleUnitId);
                 var storageAccountManager = new StorageAccountManager();
                 await storageAccountManager.CleanStorageAccount();
+                Console.WriteLine("Done");
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"An exception occured while cleaning up storage: {ex.Message}");
```
AndreasHassing, line 40: You should just use `{ex}` to get the stack trace and message.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
```diff
@@ -18,6 +18,8 @@ public WebConfig()
             webConfigPath = scaleUnit.WebConfigPath();
             configEncryptorExePath = scaleUnit.ConfigEncryptorExePath();
 
+            Decrypt();
```
AndreasHassing, line 21: We only want to do this when we need to read encrypted values, so we don't break the environment in case of errors after decrypting.

Also, decrypt a copied version of the file and use that instead.

AndreasHassing (changes requested):

Antonstockmarr:
Where should this temporary directory be located? Would it be fine to create a temporary folder in the folder with the executables for the program?

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
```diff
@@ -63,13 +62,48 @@ public void UpdateXElementIfExists(string key, string value)
...
 
+        private XElement GetDecryptedElementWithAttribute(string key)
+        {
+            string decryptPath = Path.GetTempFileName();
+            XElement element = null;
+
+            try
+            {
+                File.Copy(webConfigPath, decryptPath, true);
+                var ce = new CommandExecutor(configEncryptorExePath, "-decrypt " + decryptPath);
```
AndreasHassing, line 86: Will this work if `decryptPath` has spaces in it (i.e. if the local machine user name has spaces)? Normally in CMD and PWSH we need to enclose paths with spaces in double quotes for it to work.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
```diff
@@ -63,13 +62,48 @@ public void UpdateXElementIfExists(string key, string value)
...
         }
 
+        private XElement GetDecryptedElementWithAttribute(string key)
+        {
+            string decryptPath = Path.GetTempFileName();
+            XElement element = null;
+
+            try
+            {
+                File.Copy(webConfigPath, decryptPath, true);
```
AndreasHassing, line 85: Consider using labels for boolean parameters as it's not clear from the call-site what the `true` means.

Also, in APIs like `File.Copy` where it takes two parameters and it's unclear from the call-site if the source should be the left or right argument, consider using labels for those too.

```csharp
File.Copy(sourceFileName: webConfigPath, destFileName: decryptPath, overwrite: true);
```

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
```diff
@@ -63,13 +62,48 @@ public void UpdateXElementIfExists(string key, string value)
...
+                File.Delete(decryptPath);
+            }
+
+            return element;
+        }
+
+        private XElement GetElementWithAttribute(XDocument document, string key)
+        {
+            return document.Descendants()
+                .Where(x => (string)x.Attribute("key") == key)
```
AndreasHassing, line 103: Consider making this comparison case insensitive.

AndreasHassing (approved):
Approving with nit comments 👍.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
```diff
@@ -63,13 +62,48 @@ public void UpdateXElementIfExists(string key, string value)
 
         public string GetXElementValue(string key)
         {
-            var element = doc.Descendants()
-                .Where(x => (string)x.Attribute("key") == key)
-                .FirstOrDefault();
+            XElement element;
+            if (key.Equals("AzureStorage.StorageConnectionString"))
+            {
+                element = GetDecryptedElementWithAttribute(key);
```
Omaaarz, line 68: I would let the caller decide whether he needs a decrypted value or not, It will be easier if we had more decrypted fields that are needed from the web config. So, I'm suggesting having another method for getting values from decrypted fields

Omaaarz (commented):
Looks good to me. I have 1 suggestion.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
Antonstockmarr: Okay. My idea was to encapsulate all decryption related in the web.config. But if the caller decides whether to decrypt or not, would it not be better to have a constructor overload, decrypting the file right away and keeping the decrypted copy? That way the file is not decrypted/deleted everytime a decrypted key is needed.

Omaaarz: Love that!

AndreasHassing: In my mind, I think it's OK that the caller knows certain values have special-case handling (in this case: decryption).

For now, we don't need any further abstractions to make it easier for consumers of the `WebConfig` API to get decrypted values, but if we have more uses in the future, we can refactor based on our needs at that point in time 😊.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
Antonstockmarr: No, you're right, this will fail if the decryptPath has spaces...

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
Antonstockmarr: @AndreasHassing just to clarify, do you recommend no further action on this before it is needed?

AndreasHassing: Sorry for not being clearer.

I agree with what Omar is saying: give the caller a `GetDecryptedXElementValue` API - they will know when to use it (for `StorageConnectionString`, for example) - don't internalize knowledge about which fields are encrypted in `WebConfig.cs`; then later, if we need many fields decrypted, we can reevaluate performance and readability of the solution at that point.

I also like the constructor approach (and so did Omar) - so pick whichever 😊.

src/ScaleUnitManagement/ScaleUnitFeatureManager/Utilities/WebConfig.cs
```diff
@@ -100,7 +101,7 @@ private XElement GetDecryptedElementWithAttribute(string key)
         private XElement GetElementWithAttribute(XDocument document, string key)
         {
             return document.Descendants()
-                .Where(x => (string)x.Attribute("key") == key)
+                .Where(x => x.Attribute("key") is null ? false : ((string)x.Attribute("key")).ToLower().Equals(key.ToLower()))
```
AndreasHassing, line 104: Or with [`StringComparison`](https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=net-5.0#System_String_Equals_System_String_System_StringComparison_):
```csharp
(x.Attribute("key") as string)?.Equals(key, StringComparison.OrdinalIgnoreCase)
```

AndreasHassing: Remember, for nits: Feel free to resolve this if you believe the PR does not improve with this change.

Antonstockmarr: The problem with "?" is that "Where()" will not accept "bool?". So in case the attribute is null it must be explicitly set to false.

AndreasHassing: Aah of course. Thanks for letting me know, you're absolutely right.

Antonstockmarr: No problem :). I will still use the StringComparison to ignore case I think.

AndreasHassing (approved):

PR #102, opened 2021-10-01
Allow user to choose a subset of steps to run

### LINKED ISSUE #75: Ability to run individual steps or a subset of steps during scale unit configuration
For example: to avoid having to run DbSync while only wanting to invoke the configure endpoint in the AOS.

[Sharprompt](https://github.com/shibayan/Sharprompt) looks like it could be a good solution for this problem.

Antonstockmarr:
Create two different types of screens, the single select and the multi select used for selecting steps.
Tests included for the two screen types.
#minor
Resolves #75

### DIFF
--- a/src/CLI/EnableScaleUnitFeature.cs
+++ b/src/CLI/EnableScaleUnitFeature.cs
@@ -15,7 +15,7 @@ namespace CLI
         public override async Task Show(int input, string selectionHistory)
         {
             List<CLIOption> options = SelectScaleUnitOptions(GetSortedScaleUnits(), PrintAvailableStepsForScaleUnit);
-            var screen = new CLIScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to configure?: ");
+            var screen = new SingleSelectScreen(options, selectionHistory, "Environments:\n", "\nWhich environment would you like to configure?: ");
             await CLIController.ShowScreen(screen);
         }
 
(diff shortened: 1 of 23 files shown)


src/CLI/EnableScaleUnitFeature.cs
```diff
@@ -42,33 +42,31 @@ protected async Task PrintAvailableSteps(int input, string selectionHistory)
...
             foreach (IStep step in availableSteps)
             {
-                options.Add(new CLIOption() { Name = step.Label(), Command = RunStepsFromTask });
+                options.Add(new CLIOption() { Name = step.Label(), Command = RunStep });
             }
 
             ScaleUnitInstance scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
-            var screen = new CLIScreen(options, selectionHistory, $"Task sequence to run for {scaleUnit.PrintableName()}\n", "\nSelect task to start from: ");
+
+            string helpMessage = $"Please select steps you would like to run by supplying the step numbers\n" +
```
AndreasHassing, line 50: Should also explain that no steps chosen means: "run it all".

Wondering if we should provide what was previously there: ability for open-ended ranges.

Previously, a user would supply `3` (for instance) and then the tool would run `3-N`.

Consider if this makes sense (@Omaaarz to provide input on whether this makes sense to support, still).

AndreasHassing (commented):
Leaving for @Omaaarz to approve this, just added a few comments.

src/CLI/EnableScaleUnitFeature.cs
Antonstockmarr: Right, no input should mean run all. One could also open up for input of type "3-7" or similar.

src/CLIFramework/MultiSelectScreen.cs
```diff
@@ -100,20 +103,25 @@ private void PartitionBySign(int[] enteredNumbers, List<int> positives, List<int
...
+        private List<int> ParseInputList(string[] commaSeparatedInput)
         {
-            int[] numbers = new int[commaSeparatedInput.Length];
+            var dashSeparatedNumbers = new Regex(@"^(\d+)\s*-\s*(\d+)");
+
+            var numbers = new List<int>();
             for (int i = 0; i < commaSeparatedInput.Length; i++)
             {
                 string substring = commaSeparatedInput[i];
-                substring.Trim();
```
AndreasHassing: Why don't we want the `Trim()` anymore? I would rather that we do not fail because the user entered:
`1-2, 4,5- 7`.

src/CLI/EnableScaleUnitFeature.cs
AndreasHassing: Are we informing the user that; no input means: run all?

AndreasHassing (changes requested):

src/CLIFramework/MultiSelectScreen.cs
Antonstockmarr: I looked up the implementation of int.TryParse(), and it does not care about surrounding white spaces, so trimming actually does not change anything. The sequence you wrote there would still succeed, and I think the tests reflect this. At least the "1 - 2" case with spaces on both sides.

AndreasHassing: Gotcha, thanks 😊.

src/CLI/EnableScaleUnitFeature.cs
Antonstockmarr: You're right. I will add that

AndreasHassing (approved):

AndreasHassing:
I'd say this is a feature add, so am changing #patch to #minor.

PR #103, opened 2021-10-01
Enable single execution deployment

Antonstockmarr:
Allows the user to run the tool with argument "--deploy", which runs all steps required for setting up hub and spoke. Also allows "--clean-storage", which will clean all Azure storage accounts. If no arguments are given, the program will run as usual.

### DIFF
--- a/src/CLI/ArgumentParser.cs
+++ b/src/CLI/ArgumentParser.cs
@@ -1,0 +1,44 @@
+using System;
+
+namespace CLI
+{
+    class ArgumentParser
+    {
+        public bool deploy = false;
+        public bool cleanStorage = false;
+        private bool[] matched;
+        private string[] arguments;
+
+        public void Parse(string[] args)
+        {
+            arguments = args;
+            matched = new bool[args.Length];
+            for (int i = 0; i < matched.Length; i++)
+            {
+                matched[i] = false;
+            }
+
+            deploy = ParseArgument("--deploy");
+            cleanStorage = ParseArgument("--clean-storage");
+
+            for (int i = 0; i < matched.Length; i++)
+            {
+                if (matched[i] == false)
+                {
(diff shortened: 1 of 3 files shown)


src/CLI/ArgumentParser.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+
+namespace CLI
+{
+    class ArgumentParser
+    {
+        public bool deploy = false;
```
Omaaarz, line 7: Nit: public fields should be in PascalCase

src/CLI/Deployer.cs
```diff
@@ -0,0 +1,120 @@
...
+                Console.WriteLine($"\nCleaning environment on {scaleUnit.PrintableName()}");
+                using var context = ScaleUnitContext.CreateContext(scaleUnit.ScaleUnitId);
+                var storageAccountManager = new StorageAccountManager();
+                await storageAccountManager.CleanStorageAccount();
+            }
+        }
+
+        private async Task InitializeEnvironments()
+        {
+            foreach (ScaleUnitInstance scaleUnit in Config.ScaleUnitInstances())
```
Omaaarz, line 42: We need them sorted to initialize hub before spoke. The same goes for all other operations (prepare and install workloads)

src/CLI/Deployer.cs
```diff
@@ -0,0 +1,129 @@
...
+namespace CLI
+{
+    internal class Deployer
+    {
+        private readonly List<ScaleUnitInstance> sortedScaleUnitInstances;
+
+        public Deployer()
+        {
+            sortedScaleUnitInstances = Config.ScaleUnitInstances();
+            sortedScaleUnitInstances.Sort((s1, s2) => s1.ScaleUnitId.CompareTo(s2.ScaleUnitId));
```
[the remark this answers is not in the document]
Antonstockmarr: Ahh, I missed that it already had one.

PR #110, opened 2021-10-26
Improve error handling from AOS requests

### LINKED ISSUE #105: Show brief details about errors in ReliableRun
Currently, we log all errors to the errors.log-file - but users only see the last error in the tool when it fails after `N` retries.

This can lead to confusion as we may end up showing a transient issue unrelated to the first error. We should catch well-known errors and give detailed descriptions as to why they can occur and how to fix it (for instance: "The provided Dynamics.AX.Application.ScaleUnitEnvironmentConfiguration is invalid." could mean that the Hub was not configured while trying to configure the Spoke).

At least, we should show the error message / returned HTTP text (excluding call stack and other noisy details) on each error in retries.

Antonstockmarr:
ReliableRun class will catch AOS exceptions and try to suggest changes. Other errors pass through. If a suggestion is found, or it does not make sense to retry, the request is not repeated. Every AOS error is shown to the user.

Example of an error with an inline TSG:
image

resolves #105
#patch


AndreasHassing (approved):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/ReliableRun.cs
```diff
@@ -23,21 +24,67 @@ public static async Task Execute(Func<Task> command, string commandName)
                     await command();
                     break;
                 }
-                catch (Exception e)
+                catch (AOSClientError e)
```
AndreasHassing, line 27: Timeout exceptions and similar will not be `AOSClientError` (you only throw such an exception type if `IsSuccessStatusCode` is false - if `httpClient.SendAsync` or `response.Content.ReadAsStringAsync` throws, you won't catch it here and retry.

AndreasHassing (changes requested):

AndreasHassing:
@Antonstockmarr this PR is ready except that it needs to uptake `main`, please merge `main` into this and ensure things look good - then ping @Omaaarz or me and we can get it in 😊. Thanks!

AndreasHassing (approved):

Omaaarz (approved):

PR #115, opened 2021-11-05
Delete and truncate problematic tables #patch

### LINKED ISSUE #113: Truncate NUMBERSEQUENCEREFERENCE table during initialization
Right now we are truncating NumberSequenceScope, but not NumberSequenceReference. The two tables are linked, and they should be cleared at the same time.

Antonstockmarr:
The step to "truncate potentially problematic tables" left out tables that might cause issues. Only the NumberSequenceReference tables was truncated. NumberSequenceScope and NumberSequenceTable had rows that were related, but they were not deleted.
The step to remove problematic tables is now part of scale Unit Installation. This means that NumberSequence records can no longer be created between initializing the environment and installing workloads, which could cause the SYS package to fail.
This has been tested by running full deployment on SingleOnebox.
Resolves #113


src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -37,6 +38,60 @@ public async Task InstallationStatus()
...
+            }
+        }
+
+        private void ClearPotentiallyProblematicTables()
+        {
+            string sqlQuery = $@"
+            USE {scaleUnit.AxDbName};
+            EXEC sys.sp_set_session_context @key = N'ActiveScaleUnitId', @value = '';
+
+            DELETE FROM SysFeatureStateV0;
```
AndreasHassing, line 79: [nit] Just use `TRUNCATE` for all of these.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -37,6 +38,60 @@ public async Task InstallationStatus()
...
+            DELETE FROM FeatureManagementState;
+            DELETE FROM FeatureManagementMetadata;
+            DELETE FROM SysFlighting;
+
+            TRUNCATE TABLE NumberSequenceScope;
+            TRUNCATE TABLE NumberSequenceReference;
+            TRUNCATE TABLE NumberSequenceTable;
+
+            EXEC sys.sp_set_session_context @key = N'ActiveScaleUnitId', @value = '@A';
```
AndreasHassing, line 88: Not `@A`, but whatever the scale unit's mnemonic is.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -37,6 +38,60 @@ public async Task InstallationStatus()
...
+            await ReliableRun.Execute(async () => workloadInstances = await new WorkloadInstanceManager(hubAosClient).CreateWorkloadInstances(), "Create workload instances");
+
+            foreach (WorkloadInstance workloadInstance in workloadInstances)
+            {
+                if (await WorkloadInstanceManager.IsWorkloadInstanceInReadyState(scaleUnitAosClient, workloadInstance))
+                    continue;
+
+                if (WorkloadInstanceManager.IsSYSWorkload(workloadInstance))
+                {
+                    ClearPotentiallyProblematicTables();
```
AndreasHassing, line 57: I don't know 100%, but it's probably safer to put this down right before the `WriteWorkloadInstances`, as the workload could already be in installing state with the package just having been applied (race condition where it's about to set the state to Ready), but then we truncate the data from the initial SYS application and are in a bad state.

AndreasHassing (changes requested):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
Antonstockmarr: You're right! If we try to run Install Workloads while SYS is already being installed, these lines will clear the tables and move on, leaving us in a bad state.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -76,16 +77,16 @@ private void ClearPotentiallyProblematicTables()
             USE {scaleUnit.AxDbName};
             EXEC sys.sp_set_session_context @key = N'ActiveScaleUnitId', @value = '';
 
-            DELETE FROM SysFeatureStateV0;
-            DELETE FROM FeatureManagementState;
-            DELETE FROM FeatureManagementMetadata;
-            DELETE FROM SysFlighting;
+            TRUNCATE FROM SysFeatureStateV0;
```
AndreasHassing, line 80: `TRUNCATE TABLE`

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -52,15 +52,16 @@ private async Task InstallWorkloadsOnScaleUnit()
...
                 {
                     Console.WriteLine($"Installing the {workloadInstance.VersionedWorkload.Workload.Name} workload");
                     var workloadInstanceToInstallList = new List<WorkloadInstance>() { workloadInstance };
+
+                    if (WorkloadInstanceManager.IsSYSWorkload(workloadInstance))
+                    {
+                        ClearPotentiallyProblematicTables();
+                    }
+
                     await scaleUnitAosClient.WriteWorkloadInstances(workloadInstanceToInstallList);
```
AndreasHassing, line 65: Consider putting it immediately after the `WriteWorkloadInstances` - I have a hunch that this call in itself can add those bad records.

AndreasHassing (changes requested):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
Antonstockmarr: Hm, but I thought the ``WriteWorkloadInstances`` required these tables to be cleared first. Isn't it risky to first wait for this call to go through and then clear the tables? That could also put us in a bad state. If the query to clear the table can create new entries on its own, how can we ever assume that we have cleared these tables?

AndreasHassing: Actually, it's the first SYSHTS upload session on the scale unit that fails when these tables are not cleared first; that upload session is applied some time after `WriteWorkloadInstances` has already succeeded the first time.

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -37,6 +38,61 @@ public async Task InstallationStatus()
...
+                    continue;
+
+                if (!await WorkloadInstanceManager.IsWorkloadInstanceInInstallingState(scaleUnitAosClient, workloadInstance))
+                {
+                    Console.WriteLine($"Installing the {workloadInstance.VersionedWorkload.Workload.Name} workload");
+                    var workloadInstanceToInstallList = new List<WorkloadInstance>() { workloadInstance };
+
+                    await scaleUnitAosClient.WriteWorkloadInstances(workloadInstanceToInstallList);
+
+                    if (WorkloadInstanceManager.IsSYSWorkload(workloadInstance))
```
Omaaarz, line 62: We can have multiple SYS workload installed and we don't want to clean up these tables on the second SYS workload installation

Omaaarz (changes requested):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
Antonstockmarr: We should make sure that there are no running workloads on the scale unit before clearing the tables.

AndreasHassing: Good catch!

src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -61,7 +61,10 @@ private async Task InstallWorkloadsOnScaleUnit()
 
                     if (WorkloadInstanceManager.IsSYSWorkload(workloadInstance))
                     {
-                        ClearPotentiallyProblematicTables();
+                        if (await CountInstalledWorkloads(scaleUnitAosClient) == 0)
```
Omaaarz, line 64: I believe we should be getting the # of installed workloads before writing the workload instances save it to a variable and do only the check here.

Omaaarz (approved):

Antonstockmarr:
Tested using the single execution deploy option, everything works as expected

PR #117, opened 2021-11-09
Set the blob name in sas uri

### LINKED ISSUE #116: Importing workloads does not work with SAS token created by AX.
The SAS url created by AX is targeting the container that has the workload json file in it. But the option to import workloads in the devTool expects a SAS url for the actual current.json file. So trying to use the SAS url generated by AX will result in an access exception. The DevTools should expect a SAS url for the container, not the file in it.

Antonstockmarr:
When the sas uri is created for the container the blob name is not specified. By setting the blob name, the sas uri can still be used to copy the blob.

#patch
resolves #116

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/StorageAccountManager.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/StorageAccountManager.cs
@@ -18,7 +18,7 @@ namespace ScaleUnitManagement.ScaleUnitFeatureManager.Common
 
         public StorageAccountManager()
         {
-            var scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
+            ScaleUnitInstance scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
             connectionString = scaleUnit.AzureStorageConnectionString;
             if (string.IsNullOrEmpty(connectionString) && scaleUnit.EnvironmentType == EnvironmentType.LCSHosted)
             {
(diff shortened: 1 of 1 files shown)


Antonstockmarr:
Tested by creating a SAS url from the Azure portal for the container and importing it in the devTools.

AndreasHassing:
Why do we assume the SAS URL is created for the container, and not the blob? I would assume that we should just create a SAS URL directly for the blob?

Antonstockmarr:
There is a Geneva Action creating a SAS token for an environment, and the generated token points at the container and not the blob. But if you enter a sas for the blob it would still work.

AndreasHassing:
> There is a Geneva Action creating a SAS token for an environment, and the generated token points at the container and not the blob. But if you enter a sas for the blob it would still work.

Ah, so that's how customers/devs will get the workload instances for now, we pass a read-only SAS URL for the contanier wherein the `current.json` lives of the Hub environment?

src/ScaleUnitManagement/WorkloadSetupOrchestrator/StorageAccountManager.cs
```diff
@@ -47,12 +47,19 @@ public async Task ImportWorkloadsBlob(string sasUrlString)
...
 
-            var blobContainerName = "sysworkloadinstancesharedserviceunitstorage";
-            var blobName = "current.json";
+            string blobContainerName = "sysworkloadinstancesharedserviceunitstorage";
+            string blobName = "current.json";
 
             var targetBlobClient = new BlobClient(connectionString, blobContainerName, blobName);
 
-            var operation = await targetBlobClient.StartCopyFromUriAsync(sasUri);
+            var blobUriBuilder = new BlobUriBuilder(sasUri)
```
AndreasHassing, line 55: Will this continue to work if `sasUri` is in fact a SAS URI that already targets a blob?

AndreasHassing (approved):

src/ScaleUnitManagement/WorkloadSetupOrchestrator/StorageAccountManager.cs
Omaaarz: Yes it will. Anton mentioned that he tried passing both a SAS url for a blob and also for a container.

Omaaarz (approved):

PR #124, opened 2021-11-23
Ping AOS whenever an AOSClient is constructed

### LINKED ISSUE #106: Ping AOS before making requests
The [`AOSClient`](https://github.com/microsoft/SCMScaleUnitDevTools/blob/main/src/ScaleUnitManagement/WorkloadSetupOrchestrator/Utilities/AOSClient.cs) currently retries its actions on timeout - which is fine for the most part. But instead of showing the user a confusing error message like:
```log
System.Threading.Tasks.TaskCanceledException: The operation was canceled.
---> System.IO.IOException: Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request..
---> System.Net.Sockets.SocketException (995): The I/O operation has been aborted because of either a thread exit or an application request.
```
we should ping the AOS first to ensure it is warm before trying to perform ALM requests.

I'm not sure, but I expect an authenticated app can ping the AOS at the `POST /ping` endpoint and expect a `200 - OK` to mean that the AOS is up and running.

Antonstockmarr:
Make sure that the AOS is warm whenever a new client is created.  Improve readability of errors caused by timeout.
#patch
resolves #106
image

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/AOSCommunicator.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/AOSCommunicator.cs
@@ -20,7 +20,7 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator
             if (scaleUnitAosClient is null)
             {
                 ScaleUnitInstance scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
-                SetScaleUnitAosClient(await AOSClient.Construct(scaleUnit));
+                await ReliableRun.Execute(async () => SetScaleUnitAosClient(await AOSClient.Construct(scaleUnit)), "Connecting to AOS");
             }
             return scaleUnitAosClient;
         }
@@ -30,7 +30,7 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator
             if (hubAosClient is null)
             {
                 ScaleUnitInstance hub = Config.HubScaleUnit();
-                SetHubAosClient(await AOSClient.Construct(hub));
+                await ReliableRun.Execute(async () => SetHubAosClient(await AOSClient.Construct(hub)), "Connecting to AOS");
             }
             return hubAosClient;
         }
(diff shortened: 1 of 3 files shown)


src/ScaleUnitManagement/WorkloadSetupOrchestrator/AOSCommunicator.cs
```diff
@@ -20,7 +20,7 @@ protected async Task<IAOSClient> GetScaleUnitAosClient()
             if (scaleUnitAosClient is null)
             {
                 ScaleUnitInstance scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
-                SetScaleUnitAosClient(await AOSClient.Construct(scaleUnit));
+                await ReliableRun.Execute(async () => SetScaleUnitAosClient(await AOSClient.Construct(scaleUnit)), "Connecting to AOS");
```
Omaaarz, line 23: This will only help in the first call to AOS. all subsequent AOS calls during the life time of the tool won't get affected by the additional ping on the construction. Since we are using the same client during the life time of the tool.

Antonstockmarr: That is actually not the case right now. The AOSCommunicator does not have static clients, they are created over and over whenever we want to perform a new task in the tool. Each Client has it's own endpoint, and these are taken from the scale unit context upon creation. This is not optimal, httpclients are not meant to be recreated like this, but that is how we do it at the moment.

Omaaarz: Thanks for clarification

AndreasHassing:
@Antonstockmarr is this good to go - tested and validated?

Antonstockmarr:
> @Antonstockmarr is this good to go - tested and validated?

Yes, tested, validated and ready to go in.

This change will make sure that the AOS is responsive before making any requests, and if the connection times out, the error will say so explicitly.

AndreasHassing (approved):
LGTM! :shipit:

PR #128, opened 2022-01-07
Install workload in scale unit even if found Ready

### LINKED ISSUE #126: Workload installation should occur even when workload is Ready
To ensure upgrades are possible even without draining.

This is also in response to an internal bug where workload instances are started when installing a unrelated workload instance, making upgrade more difficult.

AndreasHassing:
We want to install workloads when they are in Ready state in Scale unit, otherwise upgrades and movement to spoke can cause problems. Since there is no _real_ workload versioning in the dev tools, we need to just overwrite on every install as we can't know what has changed without introspecting the hub workload versus what is installed in the scale unit.

#patch

Fixes #126.

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
@@ -51,9 +51,6 @@ namespace ScaleUnitManagement.WorkloadSetupOrchestrator
 
             foreach (WorkloadInstance workloadInstance in workloadInstances)
             {
-                if (await WorkloadInstanceManager.IsWorkloadInstanceInReadyState(scaleUnitAosClient, workloadInstance))
-                    continue;
-
                 if (!await WorkloadInstanceManager.IsWorkloadInstanceInInstallingState(scaleUnitAosClient, workloadInstance))
                 {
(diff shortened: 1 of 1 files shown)


src/ScaleUnitManagement/WorkloadSetupOrchestrator/ScaleUnitWorkloadInstaller.cs
```diff
@@ -51,9 +51,6 @@ private async Task InstallWorkloadsOnScaleUnit()
 
             foreach (WorkloadInstance workloadInstance in workloadInstances)
             {
-                if (await WorkloadInstanceManager.IsWorkloadInstanceInReadyState(scaleUnitAosClient, workloadInstance))
```
Omaaarz, line 54: I guess one of the reasons we might have had this reason, is that if in the previous run couple of workloads were installed, we just need to install only the remaining ones.

Not sure if that will cause us problems?

AndreasHassing: As long as a single run of workload installation succeeds, we're fine.

The new behavior will act as a workload upgrade; can that cause problems, you think?

Omaaarz: On a second thought, it shouldn't be causing problems. At least, this will save us the workarounds we are doing now.

Omaaarz (approved):

PR #131, opened 2022-01-11
Turn auto-close off during initialization #patch

### LINKED ISSUE #129: Spoke DB autoClose option should be set to False
We should be setting AutoClose option to false in scale unit DB if it was set to True. Sometimes, the spoke DB goes to "Recovery Pending" mode if that option is set to true, making the spoke environment not usable

Antonstockmarr:
If auto-close is enabled on the spoke, it will not be able to operate properly. By turning it off during initialization this is avoided.

Tested on one-box environment. Verified that the auto-close property is turned off during initialization of the spoke.

closes #129

### DIFF
--- a/src/ScaleUnitManagement/ScaleUnitFeatureManager/ScaleUnit/SetAutoClose.cs
+++ b/src/ScaleUnitManagement/ScaleUnitFeatureManager/ScaleUnit/SetAutoClose.cs
@@ -1,0 +1,30 @@
+using System.Threading.Tasks;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
+using ScaleUnitManagement.Utilities;
+
+namespace ScaleUnitManagement.ScaleUnitFeatureManager.ScaleUnit
+{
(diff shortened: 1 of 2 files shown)


src/ScaleUnitManagement/ScaleUnitFeatureManager/ScaleUnit/SetAutoClose.cs
```diff
@@ -0,0 +1,30 @@
+using System.Threading.Tasks;
+using ScaleUnitManagement.ScaleUnitFeatureManager.Utilities;
+using ScaleUnitManagement.Utilities;
+
+namespace ScaleUnitManagement.ScaleUnitFeatureManager.ScaleUnit
+{
+    public class SetAutoClose : IScaleUnitStep
```
AndreasHassing, line 7: [nit] Consider naming `DisableSqlAutoClose` or some more specific naming 😊.

AndreasHassing (approved):
Approving with suggestions 👍.

PR #132, opened 2022-01-11
Print more details during workload movement #patch

### LINKED ISSUE #125: Inform user that no movements are performed when workloads are already moved.
Movement of a workload is only done if the workload is not already on the environment it is being moved to.
If all workloads are already moved, the action will not perform any work, but it still output "done".
It should be clear to the user that the workloads were already moved, and no new movements were initiated.

Antonstockmarr:
Print the name of every workload moved when initiating movements. Also inform the user if no workloads are moved, referring the user to see movement status.

closes #125

### DIFF
--- a/src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
+++ b/src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using CloudAndEdgeLibs.Contracts;
 using ScaleUnitManagement.WorkloadSetupOrchestrator.Utilities;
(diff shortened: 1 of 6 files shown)


src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
```diff
@@ -18,6 +18,12 @@ public async Task MoveWorkloads(string moveToId)
             List<WorkloadInstance> workloadInstances = null;
```
AndreasHassing, line 18: [nit] I wonder, why does `MoveWorkloads` even take a `string moveToId` parameter? It's always the hub. Consider changing that.

AndreasHassing (approved):

AndreasHassing (approved):

PR #137, opened 2022-01-25
Set up CI with Azure Pipelines

faix:


azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
+# Starter pipeline
```
AndreasHassing, line 1: [nit] Let's remove this 😊.

azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
...
+
+trigger:
+- main
+
+pool:
+  vmImage: 'windows-latest'
+
+variables:
+  solution: '**/*.sln'
+  buildPlatform: 'Any CPU'
```
AndreasHassing, line 14: x64, as you said

build/.pipelines/azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
+# Starter pipeline
```
AndreasHassing, line 1: Why do we have 2 separate files?

AndreasHassing (changes requested):

build/.pipelines/azure-pipelines.yml
faix: just because of the way it gets created. I don't actually want it in the root folder. I will address this in an upcoming PR.

azure-pipelines.yml
faix: can't do that just yet, will address it in a future PR.

azure-pipelines.yml
faix: done

build/.pipelines/azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
...
+  buildPlatform: 'Any CPU'
+  buildConfiguration: 'Release'
+
+steps:
+- task: DotNetCoreCLI@2
+  inputs:
+    command: 'restore'
+    projects: '$(build.sourcesdirectory)/src'
+
+- task: DotNetCoreCLI@2
```
AndreasHassing, line 23: Can we give these tasks good names; like ".NET Restore", or some such?

faix: sure

AndreasHassing (approved):

PR #141, opened 2022-01-27
Don't reset the blob name on import if it has been set

vlgrozma:


AndreasHassing (approved):
Backwards compatible! I like it 👍.

vlgrozma:
> Backwards compatible! I like it 👍.

There seems to be a real reason for the old behavior: #117

AndreasHassing:
> > Backwards compatible! I like it 👍.
> 
> There seems to be a real reason for the old behavior: #117

Great catch! I had completely forgotten about that. Thanks for noticing.

PR #143, opened 2022-02-01
Automate pipeline management

Antonstockmarr:
Allow the user to drain and start data pipelines without using the UI. This is done with command line arguments "--drain-pipelines" and "--start-pipelines".

This PR also introduces actions in the CLI module. This means that the logic is separated from the menu objects used by the CLI, making it easier to automate other actions in the future.

All menu's and actions have been tested manually on a single box environment.


Antonstockmarr:
Closing this PR, because the pipeline does not allow PR's from forked repos.

See #147 for new PR.

PR #145, opened 2022-02-07
Enable updating the allowlisted apps after importing a db (#135)

faix:
* Change setup step to update table instead of inserting

The AADApp step can now be used in isolation after importing a database to update the whitelisted apps, both for scale units and the interAOS app.

* Rename whitelisting to applylisting

Complying with diversity policies


src/ScaleUnitManagement/DatabaseManager/AADAppAllowListing.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class AADAppAllowListing
+    {
+        public void UpdateAADAppClientTable(string dbName, string userName, string appName, string appId)
+        {
+            string sqlQuery = $@"
+            USE {dbName};
+
+            IF NOT EXISTS (SELECT TOP 1 1 FROM SysAADClientTable WHERE AADClientId = '{appId}')
+                BEGIN
+                    IF NOT EXISTS (SELECT TOP 1 1 FROM SysAADClientTable WHERE NAME = '{appName}')
```
AndreasHassing, line 14: The name of the SysAADClientTable row doesn't matter. That's just for labeling. Only care about the AADClientId and UserId.

src/ScaleUnitManagement/DatabaseManager/AADAppAllowListing.cs
```diff
@@ -0,0 +1,41 @@
...
+                        BEGIN
+                            IF EXISTS (SELECT TOP 1 1 FROM USERINFO WHERE ID = '{userName}')
+                                INSERT INTO SysAADClientTable (AADClientId, UserId, Name) VALUES ('{appId}', '{userName}', '{appName}');
+                            ELSE
+                                INSERT INTO SysAADClientTable (AADClientId, UserId, Name) VALUES ('{appId}', 'Admin', '{appName}');
+                        END
+                    ELSE
+                        BEGIN
+                            IF EXISTS (SELECT TOP 1 1 FROM USERINFO WHERE ID = '{userName}')
```
AndreasHassing, line 23: If the user ID does not exist in the DB- throw an error. We definitely do not want to "default" to `Admin` if they did it wrong.

AndreasHassing (changes requested):

src/ScaleUnitManagement/DatabaseManager/AADAppAllowListing.cs
Antonstockmarr: That is actually how we have done it until now. If the user "ScaleUnitManagement" did not exist, we would put in admin. But when you put it like this, I guess it is about time it changed :).

src/ScaleUnitManagement/ScaleUnitFeatureManager/Hub/AddToolToHubSysAADClientTable.cs
```diff
@@ -22,14 +23,22 @@ public Task Run()
...
-            allowListing.UpdateAADAppClientTable(hubDb, UserName, ScaleUnitAppName, scaleUnitAppId);
+                const string ScaleUnitAppName = "Scale Unit Management Tool";
+                ScaleUnitInstance scaleUnit = Config.FindScaleUnitWithId(ScaleUnitContext.GetScaleUnitId());
+                string scaleUnitAppId = scaleUnit.AuthConfiguration.AppId;
+                allowListing.UpdateAADAppClientTable(hubDb, UserName, ScaleUnitAppName, scaleUnitAppId);
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine(ex);
+                Console.WriteLine($"\nIf the {UserName} user does not exist, run Deployment.Setup.exe fullall sync to create it");
```
AndreasHassing, line 40: Would prefer not to have the duplication between the two `AddToolTo..` components.

Maybe `UpdateAADAppClientTable` can catch the exception, inspect it and if it contains "User does not exist" in the message, we write out what the user can do about it, then throw the exception back up to fail the operation.

Antonstockmarr:
How the error will look if the user is not there
image

AndreasHassing (approved):

PR #147, opened 2022-02-08
Automate pipeline management

Antonstockmarr:
Allow the user to drain and start data pipelines without using the UI. This is done with command line arguments "--drain-pipelines" and "--start-pipelines".

This PR also introduces actions in the CLI module. This means that the logic is separated from the menu objects used by the CLI, making it easier to automate other actions in the future.

All menu's and actions have been tested manually on a single box environment.


src/CLI/IAction.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CLI
+{
+    interface IAction
```
AndreasHassing, line 8: [nit] add accessibility modifier

src/CLI/IAction.cs
```diff
@@ -0,0 +1,12 @@
...
+    {
+        public Task Execute();
```
AndreasHassing, line 10: For interfaces, visibility modifiers are not needed;
```suggestion
        Task Execute();
```

src/CLI/Menus/ConfigureEnvironment.cs
```diff
@@ -17,10 +16,9 @@ public override async Task Show(int input, string selectionHistory)
 
         private async Task ConfigureScaleUnit(int input, string selectionHistory)
         {
-            if (ScaleUnitContext.GetScaleUnitId() == "@@")
-                await new HubConfigurationManager().Configure();
-            else
-                await new ScaleUnitConfigurationManager().Configure();
+            string scaleUnitId = GetSortedScaleUnits()[input - 1].ScaleUnitId;
```
AndreasHassing, line 19: I'm betting this concept of using `GetSortedScaleUnits()[input - 1]` is used a lot to find a concrete scale unit?

If it is used more than twice, consider refactoring into a shared method on the base, such that you can call:
```csharp
string scaleUnitId = GetScaleUnitFromIndex(input).ScaleUnitId;
```

src/ScaleUnitManagementTests/ArgumentParserTest.cs
```diff
@@ -22,25 +22,29 @@ public void BeforeParsing_HasInitialValues()
...
+            argumentParser.DrainPipelines.Should().BeFalse();
+            argumentParser.StartPipelines.Should().BeFalse();
         }
 
         [TestMethod]
         public void Parse_WithValidArguments_SetsValues()
         {
             // Arrange + Act
-            string[] arguments = { "--single-box-deploy", "--clean-storage" };
+            string[] arguments = { "--single-box-deploy", "--clean-storage", "--drain-pipelines", "--start-pipelines" };
```
AndreasHassing, line 33: [observation] These tests would not be needed with DragonFruit or System.CommandLine

src/CLI/Utilities/ArgumentParser.cs
```diff
@@ -36,5 +41,15 @@ private bool ParseArgument(string argument)
             }
             return false;
         }
+
+        public string HelpMessage()
+        {
+            return "Usage: program [options]\n" +
```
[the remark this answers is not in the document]
AndreasHassing: Not in this PR, but consider creating a work item for future improvement :).

AndreasHassing (approved):

PR #150, opened 2022-02-14
Enable multibox deployment script

Antonstockmarr:
The user can type --hub-deploy or --spoke-deploy to run deployment for one of the scale units.
The deployer class is made abstract, and is implemented by the SingleBoxDeployer and the ScaleUnitDeployer.

### DIFF
--- a/src/CLI/Program.cs
+++ b/src/CLI/Program.cs
@@ -26,32 +26,14 @@ namespace CLI
             }
             catch (Exception ex)
             {
-                Console.WriteLine(ex);
+                Console.WriteLine(ex.Message);
                 Console.WriteLine("\n" + argumentParser.HelpMessage());
                 return;
             }
 
-            var deployer = new Deployer();
+            var argumentHandler = new ArgumentHandler();
+            await argumentHandler.RunScripts(argumentParser);
 
-            if (argumentParser.CleanStorage)
-            {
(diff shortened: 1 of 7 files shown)


src/CLI/Program.cs
```diff
@@ -26,32 +26,14 @@ public static async Task Main(string[] args)
             }
             catch (Exception ex)
             {
-                Console.WriteLine(ex);
+                Console.WriteLine(ex.Message);
```
AndreasHassing, line 29: Why only `Message`? We should keep printing the full exception, right? This does not give us a stack trace to work with in user-reported bugs.

AndreasHassing:
How was this tested?

Added comments, looks good.

AndreasHassing (changes requested):

Antonstockmarr:
PR has been tested on the single one box by running the hub deploy and spoke deploy and validating that the environments are correctly set up. Have also been testing the different combinations of argument inputs.

src/CLI/Program.cs
Antonstockmarr: I was thinking that the exception messages from the ArgumentParser were enough explanation, but that is probably a bold assumption. I will change it back.

AndreasHassing (approved):

PR #154, opened 2022-02-22
Use 'ScaleUnitPipeline' user for the interAOS app

### LINKED ISSUE #151: Update SysAADClientTable user ID to ScaleUnit(Management|Pipeline)
During [AAD App allowlisting](https://github.com/microsoft/SCMScaleUnitDevTools/blob/87c739c8e4143315f3fd875378913b32dfe34269/src/ScaleUnitManagement/DatabaseManager/AADAppAllowListing.cs#L16-L17), we don't update the user to `ScaleUnitPipeline` for `ScaleUnits` app.

Internally reported bug from @vlgrozma.

Antonstockmarr:
Inserting the interAOS appId into SYSAADCLIENTTABLE will use the user 'ScaleUnitPipeline' if it exists, otherwise 'Admin'.
Tested on the single-one-box environment.

closes #151


src/ScaleUnitManagement/Utilities/SqlQueryExecutor.cs
```diff
@@ -1,26 +1,60 @@
+using System;
 using System.Data.SqlClient;
 
 namespace ScaleUnitManagement.Utilities
 {
     public class SqlQueryExecutor
     {
+        private readonly string dbName;
         private readonly ScaleUnitInstance scaleUnit;
```
AndreasHassing, line 9: [nit] This no longer needs to exist as a class field.

AndreasHassing (approved):

PR #155, opened 2022-02-22
Make emergency transition independent of the spoke

### LINKED ISSUE #149: Emergency transition to hub depends on the spoke environment
When performing emergency transition to hub, the workloads are fetched from the spoke. But if the spoke environment is broken this fetch may fail or be corrupted. A broken spoke might be the reason for performing this action in the first place. The transition should depend only on the hub, and not on the spoke.

Antonstockmarr:
The emergency transition does no longer depend on the spoke. It gets the workload definitions from the hub, and sees which ones are assigned to the spoke.
Improve visibility for the user, so they can see which workloads are affected, if any.
Check if there are no EnvironmentOverrides already on the workload, since the overrides should not be set multiple times.
Tested on singleOneBox environment.

Closes #149


src/ScaleUnitManagement/WorkloadSetupOrchestrator/WorkloadMover.cs
```diff
@@ -51,15 +51,35 @@ public async Task MoveWorkloadsToHub()
...
+                    continue;
+                }
+
+                if (workloadInstance.ExecutingEnvironment.Any())
+                {
+                    TemporalAssignment lastAssignment = workloadInstance.ExecutingEnvironment.Last();
+                    if (scaleUnitId.Equals(lastAssignment?.Environment.ScaleUnitId))
+                    {
+                        workloadInstance.ExecutingEnvironmentOverride = CreateTemporalAssignment("@@", DateTime.UtcNow);
+                        await ReliableRun.Execute(async () => await hubAosClient.WriteWorkloadInstances(new List<WorkloadInstance> { workloadInstance }), $"Moving workload instance from scale unit {scaleUnit.ScaleUnitId} to the hub");
```
AndreasHassing, line 73:
```suggestion
                        await ReliableRun.Execute(async () => await hubAosClient.WriteWorkloadInstances(new List<WorkloadInstance> { workloadInstance }), $"Performing emergency transition for workload instances from scale unit {scaleUnit.ScaleUnitId} to the hub");
```

AndreasHassing (approved):
Minor comments, otherwise looks good 😊👍.

PR #156, opened 2022-03-01
Restructure menus, so they reflect the SUM portal better #minor

Antonstockmarr:
Introduce Navigation menus and Leaf menus, making a more clear distinction between menus that perform actions and menus that don't.
It has been manually tested that the menus are wired together directly.

New menu structure looks like this:
DevTools menus new


src/CLI/LeafMenu.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading.Tasks;
+
+namespace CLI
+{
+    internal abstract class LeafMenu : DevToolMenu
```
AndreasHassing, line 5: [suggestion] Consider naming `MenuAction` or `MenuLeafAction` (if you want to keep the tree-lingo), which I believe better expresses what it's for.

Are you using tree lingo anywhere else for the menu structure? I like it, just wondering if this `LeafMenu` artifact is the odd one out.

src/CLI/DevToolMenu.cs
```diff
@@ -10,6 +10,8 @@ public abstract class DevToolMenu : CLIMenu
     {
         protected List<ScaleUnitInstance> sortedScaleUnits = null;
 
+        public abstract string Description { get; }
```
AndreasHassing, line 13: [nit/preference] Consider naming `Label` or `Title` (I personally prefer label), as description is usually for texts that are a bit longer than a label or title and adds additional details to some item.

src/CLI/LeafMenu.cs
```diff
@@ -0,0 +1,9 @@
...
+    {
+        protected abstract Task PerformAction(int input, string selectionHistory);
```
AndreasHassing, line 7: What is the benefit of the `LeafMenu` abstraction?

AndreasHassing (approved):
Minor comments, good work!

src/CLI/LeafMenu.cs
Antonstockmarr: The benefit is more a semantic one of knowing when a menu is for navigation and when it is for executing actions. All of the menus had different actions that essentially did the same thing. My thinking was that this would make it easier to create new menus following the same scheme, just like all menus are expected to implement a "Show" method. Another benefit is that it would be easier in the future to autogenerate documentation of the menu structure using reflection. But it could be that I am overcomplicating things...

src/CLI/LeafMenu.cs
Antonstockmarr: There is no other tree lingo in the project. I think I need to reconsider what the purpose of this class is.

src/CLI/LeafMenu.cs
AndreasHassing: The autogenerating of documentation sounds very cool, but yes, perhaps not a very worthwhile investment at this time 😊.

It's fine to keep it, just wanted you to think and share your perspective here. Normally I wouldn't introduce an abstraction like this which doesn't reduce code duplication or has similar gains; but providing future developers with a pattern and forced scheme sounds good- too.

faix:
abandoning as we are archiving the repo

PR #157, opened 2022-03-08
Fix bug when entering wrong input

### LINKED ISSUE #133: Sometimes the input read by the CLI is out of sync
It seems that the CLI can sometimes end in a bad state. This might be caused by entering things while the terminal is processing a request or similar. In the image below, the number 62 was entered, followed by 3. The tool read 62 one cycle later than it was entered, not recognizing the 3. 
image

Antonstockmarr:
When entering a wrong input in the CLI, like a number outside the range of options, the program would never clear the input error status with subsequent inputs. The program would be stuck in a bad state. Now the error is cleared every time before validating inputs.

Tests included.
Resolves #133

### DIFF
--- a/src/CLIFramework/MultiSelectScreen.cs
+++ b/src/CLIFramework/MultiSelectScreen.cs
@@ -12,6 +12,8 @@ namespace CLIFramework
 
         public override async Task PerformAction(string input)
         {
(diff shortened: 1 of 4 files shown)


faix:
Abandoning as we archiving the repo.

